PR #1, opened 2015-10-16
Fix incorrect GII "length" field descriptions

dcommander:
The "length" field should always contain the total number of bytes in
the message, minus the type, subtype, and length fields.

I confirmed that this is the actual behavior of ggivnc, which lends
credence to the fact that this is simply a documentation error.


CendioOssman:
Thanks. Looks good, but could you add a signed-off-by in accordance with the README?

dcommander:
done

PR #4, opened 2016-08-17
Vencrypt fixes

jacmet:
Minor wording tweaks and the addition of the undocumented server side vencrypt subtype ack message

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -674,8 +674,8 @@ Code            Name            Description
 264             X509SASL        X509 encryption with SASL authentication
 =============== =============== =======================================
 
-After that client selects one VeNCrypt subtype sends back number of that
-type.
+After that client selects one VeNCrypt subtype and sends back the
+number of that type.
 
 =============== ======= ===============================================
 No. of bytes    Type    Description
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -686,6 +686,16 @@ No. of bytes    Type    Description
 If client supports none of the VeNCrypt subtypes it terminates
 connection.
 
+After that server sends one byte response which indicates if everything
+is OK. Non-one value means failure and connection will be closed. One
+value means success.
+
+=============== ======= ===============================================
+No. of bytes    Type    Description
+=============== ======= ===============================================
+1               ``U8``  Ack
+=============== ======= ===============================================
+
 When subtype is selected authentication continues as written in particular
 VeNCrypt subtype description.
 


CendioOssman:
That's not good. We have not implemented that extra ack in TigerVNC, so changing the specification at this point is going to cause issues. We can't ignore it either of course, so we need to find some way to migrate things.

jacmet:
So TigerVNC isn't compatible with the VNC server in QEMU, libvncserver/x11vnc or the gtk-vnc client if VeNCrypt is used? That's pretty bad :/

CendioOssman:
Yes, it does seem like it is that bad. :(
We probably need to tick up the VeNCrypt version number and get everyone in sync around the same behaviour.

jacmet:
I guess that is the only solution. I think it is doable to detect the extra ack  from the client in the server before the TLS handshake (as the first byte of the handshake afaik won't be '1'), but compatibility between tigervnc client and other servers seems impossible to me (unless the client is updated to also send the ack, but then it won't be compatible with older versions of the tigervnc server).

peda-r:
Please note that the ack byte seems to be sent *AFTER* the handshake (which of course is madness).

But at least that was what I had to implement in order to get my stuff to interact with the original VeNCrypt software. If I remember correctly...

My point is check this against VeNCrypt before commiting anything...

jacmet:
No, for qemu/x11vnc/gtk-vnc/libvncserver at least - The ack byte sent from the server after the client has selected a vencrypt subtype from the list provided by the server.

CendioOssman:
> But at least that was what I had to implement in order to get my stuff to interact with the original VeNCrypt software. If I remember correctly...

I had a look at the original implementation and I couldn't see any trace of that extra ack.

peda-r:
Oh crap, I remembered something was odd with that ack appearing after the handshake, and I then misread my code. It's been a while and the code looks strangely alien to me by now...

I guess I must once have had a bug that made the ack appear after the handshake or the other side expecting it after the handshake. Or something?

Anyway, I can now confirm that both my client and my server have the ack *before* the handshake.

jacmet:
Where's the source code of the original implementation?

The logic in qemu is here: http://git.qemu.org/?p=qemu.git;a=blob;f=ui/vnc-auth-vencrypt.c;h=11c8c9a819fce621eeabe6b020c8e4d90b3677f9;hb=HEAD#l86

And in libvncclient: https://github.com/LibVNC/libvncserver/blob/ddabcb67a630d5b8e73cdce6d78f9413b36b742b/libvncclient/tls_gnutls.c#L401

peda-r:
I assume it's this:
https://sourceforge.net/p/vencrypt/code/HEAD/tree/

jacmet:
Ahh yes, the original code doesn't seem to send this ack. I wonder where it comes from?

peda-r:
I have the ack in my code as well, and I'm pretty sure I tested interoperability with the original code at some point. But maybe I'm misremembering that too, and maybe I ended up failing to test that for some reason and settled for something else for interoperability testing?

Not that all the C++ goo in VeNCrypt is all that easy to follow, maybe the ack is unintentional?

peda-r:
I found this:

Client side:
https://sourceforge.net/p/vencrypt/code/HEAD/tree/tags/VeNCrypt_0_2_3/common/rfb/CSecurityTLSBase.cxx#l97

Server side:
https://sourceforge.net/p/vencrypt/code/HEAD/tree/tags/VeNCrypt_0_2_3/common/rfb/SSecurityTLSBase.cxx#l111

jacmet:
Ahh yes, there it is. It apparently has been like this since basically the beginning:

VeNCrypt 0.1.1
------------------------------------------------------------------------
r5 | sibecker | 2006-05-23 17:09:33 +0200 (Tue, 23 May 2006) | 1 line

CendioOssman:
Hah. Well that was nice and obscure. I can find the same thing in TigerVNC, fortunately. But it does show that the ack is only for TLS sub-types, and no other. So your patch needs to be tweaked a bit.

(TigerVNC also expects a reason code after a nack for some reason, even though our own server code doesn't send one)

jacmet:
You mean it isn't sent for type 256 (plain)? True, I'll update that.

CendioOssman:
Or any of the non-VeNCrypt authentication methods.

jacmet:
Ehh, are they allowed as VeNCrypt sub types? That isn't clear from rfbproto and not supported by E.G. QEMU.

peda-r:
Aha, but there are no non-VeNCrypt subtypes defined after you have opened the VeNCrypt security type, you can only select 256-264. In other words, you have to open the Tight security type first, and then VeNCrypt, the other way around is not supposed to work.

FWIW, my VeNCrypt implementations only have acks for subtypes 257-264.
I guess that means that the only thing out of sync is the spec?

jacmet:
@peda-r yes, that's my understanding as well.

CendioOssman:
You can send the other ones as well. The original implementation supports this, as does TigerVNC. I know others fail to do so though. However you cannot do opportunistic encryption without supporting the basic types after negotiating VeNCrypt.

So the spec should be updated to clarify that any auth type is valid.

jacmet:
@CendioOssman  "you cannot do opportunistic encryption without supporting the basic types". Why is that? 

What I do in my server is to offer both classic VNC auth and VeNCrypt in the initial security type list (VeNCrypt listed first, as it is preferred). If clients support VeNCrypt (and hence TLS), then they will select that and older clients fall back to VNC auth.

Why would clients support the VeNCrypt handshake and then only the older auth types? That would just be pointless indirection.

CendioOssman:
The server might not have a VeNCrypt type the client finds acceptable. It cannot know this until it has performed the VeNCrypt handshake. The cause of this might be lack of symmetry in the server or client configuration, or the fact that VeNCrypt does not have equivalents for every non-VeNCrypt authentication.

But I'd say the question is backwards. Why are you preventing this? You gain nothing by it, and you make yourself incompatible with the original VeNCrypt implementation.

jacmet:
Ok, so the desired behaviour is that after the VeNCrypt handshake the server SHOULD NOT enumerate the non-VeNCrypt security types in the list of subtypes, BUT it should accept them if supported (E.G. if it had listed them in the initial security list message). If the client doesn't support any of the listed VeNCrypt subtypes, it MAY instead select one of the other (E.G. not VeNCRypt again) security types from the initial security list.

Correct? If so, I'll try to add some text about this.

CendioOssman:
It should both enumerate and accept them.

peda-r:
Really? I find it highly questionable that a client requesting VeNCrypt can reasonably accept a fallback to any non-VeNCrypt security type. Any instruction from the user to the client software to "encrypt if you can" is just crazy, no? I mean, either you have set up encryption, and then you want to use it. Or, you have not set up encryption, and then you don't need to open the VeNCrypt security type.

(VeNCrypt plain (256) does not count, it's just plain silly with the password in clear text.)

CendioOssman:
There are other encryption methods besides VeNCrypt. And besides, that's a policy that should be left to the client.

jacmet:
Sorry for the slow response, I've adjusted the branch to:
- Clarify that ack message is only sent for TLS/X509 subtypes
- Clarify that non-vencrypt subtypes may be enumerated in the vencrypt handshake as well

CendioOssman:
I think you pushed to the wrong branch. :)

jacmet:
@CendioOssman Really? I must say this is the first time I use github pull requests instead of normal emails, but I do see them on my vencrypt branch: https://github.com/jacmet/rfbproto/commits/vencrypt and they are listed here above as well

CendioOssman:
Must be because it is closed in that case. And I can't seem to reopen it. :/
I think you'll have to make a new PR, unfortunately.

jacmet:
Ok, I've created #5 instead

PR #8, opened 2016-12-29
Clarify the existing uses of Compression Level Pseudo-encoding

patrakov:
When compression level is high, old Vino (e.g. 3.8.1) takes it as a
hint to prefer JPEG when using the Tight encoding, while modern
libvncserver thinks it should avoid JPEG and prefer palette.

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -3217,6 +3217,10 @@ levels should be used for any given bandwidth. The compression level is
 just a hint for the server, and there is no specification for what a
 specific compression level means.
 
+There are VNC server implementations that also use compression level
+as a hint to prefer or deprefer lossless compression mechanisms over
+JPEG-based ones.
+
 QEMU Pointer Motion Change Pseudo-encoding
 ------------------------------------------
 


CendioOssman:
Servers use all kinds of heuristics. I don't think we should document such behaviours unless they have a significant effect on the actual protocol.

patrakov:
Well, currently the previous paragraph leaves the misleading impression that this pseudo-encoding is relevant only for lossless compression. That's why the change. Can you suggest any other way to remove this false impression?

CendioOssman:
That's what it was defined to do. We can't really add "Some servers also use this as a hint for other things" as we'd quickly add that statement to pretty much every paragraph.

This setting implies intent, and we need to keep that intent clear so that developers can label and document it properly. If you think that Vino and libvncserver are not respecting that intent then I'd say that's a protocol violating bug that they should fix.

CendioOssman:
Although protocol violations don't belong in the spec, we could have a wiki page to document interesting cases.

patrakov:
> If you think that Vino and libvncserver are not respecting that intent then I'd say that's a protocol violating bug that they should fix.

I see your point. However, I cannot really report it that way, because then I would have to suggest an alternative that doesn't break the existing Tight clients' ability to change the compression level. Want it or not, the use of this pseudo-encoding for non-lossless stuff, as a general knob for bandwidth/cpu tradeoff, is in the de-facto protocol.

I am OK with putting it on the wiki page, or leaving documented in a closed pull request.

patrakov:
I would also be OK with dropping the word "lossless" from the previous paragraph, e.g.:

The compression level concerns the general tradoff between CPU time and bandwidth. It is therefore
probably difficult to define exact cut-off points for which compression
levels should be used for any given bandwidth. The compression level is
just a hint for the server, and there is no specification for what a
specific compression level means for each particular encoding.

CendioOssman:
> I see your point. However, I cannot really report it that way, because then I would have to suggest an alternative that doesn't break the existing Tight clients' ability to change the compression level.

How so?

> I would also be OK with dropping the word "lossless" from the previous paragraph, e.g.:

Maybe. But the point is to differentiate it from the quality hint. Perhaps a more careful wording rather than just dropping the word.

CendioOssman:
How about the following:

```rst
The compression level concerns the general tradeoff between CPU time
and bandwidth. It is therefore probably difficult to define exact cut-off
points for which compression levels should be used for any given bandwidth.
The compression level is just a hint for the server, and there is no
specification for what a specific compression level means for each
particular encoding.

Most servers use this hint to control lossless compression algorithms as
the tradeoff between CPU time and bandwidth is obvious there. However it
can also be used for other algorithms where this tradeoff is relevant.
```

patrakov:
Good wording!

CendioOssman:
Fixed in 4001074a0fd79d38dda854852e6d59c09ca00501.

PR #11, opened 2018-02-26
Document Tight PNG encoding.

kanaka:
Relevant issue: https://github.com/rfbproto/rfbproto/issues/9

The Tight PNG encoding was designed as part of a Google Summer of Code
project by Corentin Chary, Alexander Graf, Anthony Liguori, and Joel
Martin in June 2010.

The encoding was initially implemented in noVNC, QEMU and libvncserver
by the following commits:
    - https://github.com/novnc/noVNC/commit/29cb15f9caf3130260dc5d3e64235f77e21d89dc (Jun 9, 2010)
    - https://github.com/qemu/qemu/commit/efe556adb75a20ab71f3e5b1c5b19bf045e7953f (Jul 7, 2010)
    - https://github.com/LibVNC/libvncserver/commit/896ca2036c35b89a7f63e1adefe5e3724bf4d40d (Jul 19, 2011)

Relevant links:
    https://wiki.qemu.org/Features/VNC_Tight_PNG
    https://wiki.qemu.org/Google_Summer_of_Code_2010/VNC#Tight_PNG
    https://sourceforge.net/p/libvncserver/mailman/message/29200412/

Signed-off-by: Joel Martin <github@martintribe.org>

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -2315,6 +2315,14 @@ Number       Name
 7            `Tight Encoding`_
 8            `zlibhex Encoding`_
 16           `ZRLE Encoding`_
+-260         `Tight PNG Encoding`_
+============ ==========================================================
+
+The pseudo-encodings defined in this document are:
+
+============ ==========================================================
+Number       Name
+============ ==========================================================
 -23 to -32   `JPEG Quality Level Pseudo-encoding`_
 -223         `DesktopSize Pseudo-encoding`_
 -224         `LastRect Pseudo-encoding`_
@@ -2359,7 +2367,6 @@ Number                      Name
 -225                        PointerPos
 -226 to -238                Tight options
 -241 to -246                Tight options
--260                        Tight PNG
 -262 to -272                QEMU
 -273 to -304                VMWare
 -306                        popa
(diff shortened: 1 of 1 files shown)


CendioOssman:
Looks good, except for the sorting. So far "real" encodings have been positive numbers, and pseudo-encodings have been negative numbers. This unfortunately breaks that pattern. 

However I still think it is good if the list follows the chapter order. So far we've followed the number for ordering, even if there are other related encodings. So unless we have a really good reason to abandon that, I would suggest sorting it at the end of the encodings. I'm worried we might degenerate in to some arbitrary mess otherwise.

For clarity we might also want to split that table in to two and describe the second as "The pseudo-encodings defined in this document are:". It's not as easy to spot the transition now that you can't just look for negative numbers. :)

rfbproto.rst
```diff
@@ -2855,6 +2855,45 @@ compression method.
...
     **FillCompression** or **JpegCompression**.
 
+Tight PNG Encoding
+-------------------------
+
+The Tight PNG encoding is a variant of the Tight encoding that
+disallows the **BasicCompression** type and replaces it with a new
+**PngCompression** type. The purpose of this encoding is to support
+clients which have efficient PNG decoding/rendering (perhaps even in
+hardware) but may have inneficient decoding of the raw zlib data
```
peda-r: inefficient

CendioOssman:
Ping @kanaka :)

kanaka:
Wow, I really let the ball drop on that one! I had a good excuse in that I've been working on my PhD dissertation. But still, I could have spared the 10 minutes to do this update :-(.  Anyways, I've pushed the table split and the re-ordering (and fixed the silly typo, thanks @peda-r!). The one thing that is a little bit weird is that the tables for both encodings and pseudo-encodings are in the encoding section and there is a pseudo-encoding section at the same level and the encoding section. It was that way before too, it's just more obvious now with the table split.

I kept this as a separate commit. Let me know if you would like it squashed and I'll do that.

CendioOssman:
> Wow, I really let the ball drop on that one! I had a good excuse in that I've been working on my PhD dissertation. But still, I could have spared the 10 minutes to do this update :-(. Anyways, I've pushed the table split and the re-ordering (and fixed the silly typo, thanks @peda-r!). The one thing that is a little bit weird is that the tables for both encodings and pseudo-encodings are in the encoding section and there is a pseudo-encoding section at the same level and the encoding section. It was that way before too, it's just more obvious now with the table split.
> 

I agree. However we have the third category of "Unknown encodings", which would be unclear where they would end up. Let's leave that issue for the future. :)

> I kept this as a separate commit. Let me know if you would like it squashed and I'll do that.

I think a squash makes sense here. So please do that and I'll merge this. :)

kanaka:
Squashed.

PR #14, opened 2018-06-11
Add VMWare psuedo-encodings

qmfrederik:
This is an attempt to document the VMWare pseudo-encodings.

References
- https://wiki.multimedia.cx/index.php/VMware_Video
- http://www.ffmpeg.org/doxygen/0.6/vmnc_8c-source.html#l00287

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1016,7 +1016,7 @@ Number      Name
 20          RequestSession
 21          SetSession
 80          NotifyPluginStreaming
-127         VMWare
+127         VMware
 128         Car Connectivity
 150         `EnableContinuousUpdates`_
 248         `ClientFence`_
@@ -1025,7 +1025,7 @@ Number      Name
 251         `SetDesktopSize`_
 252         tight
 253         `gii Client Message`_
-254         VMWare
+254         VMware
 255         `QEMU Client Message`_
 =========== ===========================================================
 
@@ -1995,7 +1995,7 @@ Number      Name
 13          KeepAlive
 14          Possibly used in UltraVNC
 15          ResizeFrameBuffer
-127         VMWare
+127         VMware
 128         Car Connectivity
 150         `EndOfContinuousUpdates`_
 173         ServerState
@@ -2004,7 +2004,7 @@ Number      Name
 250         `xvp Server Message`_
 252         tight
 253         `gii Server Message`_
-254         VMWare
+254         VMware
 255         `QEMU Server Message`_
 =========== ===========================================================
 
@@ -2342,6 +2342,13 @@ Number       Name
 -314         `Cursor With Alpha Pseudo-encoding`_
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
 -763 to -768 `JPEG Subsampling Level Pseudo-encoding`_
+0x574d5664   `VMware Cursor Pseudo-encoding`_
+0x574d5665   `VMware Cursor State Pseudo-encoding`_
+0x574d5666   `VMware Cursor Position Pseudo-encoding`_
+0x574d5667   `VMware Key Repeat Pseudo-encoding`_
+0x574d5668   `VMware LED state Pseudo-encoding`_
+0x574d5669   `VMware Display Mode Change Pseudo-encoding`_
+0x574d566a   `VMware Virtual Machine State Pseudo-encoding`_
 0xc0a1e5ce   `Extended Clipboard Pseudo-encoding`_
 ============ ==========================================================
 
@@ -2369,13 +2376,13 @@ Number                      Name
 -241 to -246                Tight options
 -260                        Tight PNG
 -262 to -272                QEMU
--273 to -304                VMWare
+-273 to -304                VMware
 -306                        popa
 -310                        OLIVE Call Control
 -311                        ClientRedirect
 -523 to -528                Car Connectivity
 0x48323634                  VA H.264
-0x574d5600 to 0x574d56ff    VMWare
+0x574d5600 to 0x574d56ff    VMware
 0xc0a1e5cf                  PluginStreaming
 0xfffe0000                  KeyboardLedState
 0xfffe0001                  SupportedMessages
@@ -3606,6 +3613,205 @@ The values for this pseudo-encoding are defined as follows:
 
 This pseudo-encoding was originally intended for use with JPEG-encoded
 subrectangles, but it could be used with other types of image encoding as well.
+
+VMware Cursor Pseudo-encoding
+-----------------------------
+
+A server sets the cursor shape by sending a pseudo-rectangle with the
+VMware Cursor pseudo-encoding as part of an update. The
+pseudo-rectangle's *x-position* and *y-position* indicate the hotspot
+of the cursor, and *width* and *height* indicate the width and height
+of the cursor in pixels.
+
+The data starts with a *cursor-type*  byte followed by padding:
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+1               ``U8``               *cursor-type*
+1               ``U8``               padding
+=============== ==================== =======================
+
+The value of *cursor-type* is either 0 to indicate a 'classic' cursor
+which consists of an AND mask and a XOR mask, or 1 indicate an alpha
+cursor.
+
+The data for a classic cursor consists two sets of *width* * *height*
+pixel values, defining an AND mask and a XOR mask:
+
+======================================= =================== =============
+No. of bytes                            Type                Description
+======================================= =================== =============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *and-mask*
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *xor-mask*
+======================================= =================== =============
+
+Classic cursors can be drawn using the following code:
+
+::
+
+  dst[i] = (dst[i] & and-mask[i]) ^ xor-mask[i];
+
+The data for an alpha cursor consists of *width* * *height* RGBA pixel
+values:
+
+======================================= =================== ==============
+No. of bytes                            Type                Description
+======================================= =================== ==============
+*width* * *height* * 4                  ``U8`` array        *cursor bytes*
+======================================= =================== ==============
+
+Alpha cursors should be drawn by compositing the cursor image into the
+framebuffer.
+
+VMware Cursor State Pseudo-encoding
+-----------------------------------
+
+A server sets the cursor state by sending a pseudo-rectangle with the
+VMware Cursor State pseudo-encoding as part of an update.
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+2               ``U16``              *cursor-state*
+=============== ==================== =======================
+
+*cursor-state* is a bit-field which describes the state of the cursor:
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+0               Cursor visible.
+1               Cursor absolute.
+2               Cursor wrap. 
+=============== =======================================================
+
+If cursor visible is not set, the cursor should not be rendered into
+the framebuffer until another cursor state message is received that
+turns the cursor back on.
+
+The cursor absolute bit indicates whether the virtual machine is using
+an absolute or relative mouse.
+
+The cursor warp bit is set when the virtual machine artificially moves
+the position of the cursor. This value is for information purposes
+only.
+
+VMware Cursor Position Pseudo-encoding
+--------------------------------------
+
+A server updates the cursor position by sending a pseudo-rectangle with
+the VMware Cursor Position pseudo-encoding. The *x-position* and
+*y-position* define the new position of the cursor hot spot (not the
+the top left corner of cursor image).
+
+VMware Key Repeat Pseudo-encoding
+---------------------------------
+
+The server notifies the client of changes to the keyboard key repeat by
+sending a pseudo-rectangle with the VMware Key Repeat
+pseudo-encoding. It is used to determine whether the client or server
+should handle the key repeat.
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+2               ``U16``              *key-repeat-enabled*
+4               ``U32``              *period*
+4               ``U32``              *delay*
+=============== ==================== =======================
+
+*key-repeat-enabled* is 1 if the VNC client should handle key
+repeat, 0 if the server handles key repeat.
+*period* defines the period to wait between key repeats.
+*delay* defines the delay for the first key repeat.
+
+VMware LED State Pseudo-encoding
+---------------------------------
+
+The server sends a pseudo-rectangle with the VMware LED State
+pseudo-encoding to toggle the state of lock keys on the keyboard.
+
+This pseudo-rectangle has the following contents:
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+4               ``U32``              *led-flags*
+=============== ==================== =======================
+
+*led-flags*  is a bitmask which defines whether a keyboard LED is on or
+off.
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+0               Scroll Lock
+1               Num Lock
+2               Caps Lock
+=============== =======================================================
+
+The remaining bits are reserved and must be ignored.
+
+VMware Display Mode Change Pseudo-encoding
+------------------------------------------
+
+The server changes the desktop size by sending a pseudo-rectangle with
+the VMware Display Mode Change pseudo-encoding.
+
+This encoding is used for scenarios not addressed by the
+`DesktopSize Pseudo-encoding`_.
+
+Specifically, the `DesktopSize Pseudo-encoding`_ does not allow the VNC
+server to redefine both the color depth and size of the framebuffer.
+This is useful when the client prefers to receive the framebuffer
+native color depth at all times. It is defined to be similar to the
+ServerInitialisation header to facilitate client implementations.
+
+The new pixel format takes effect immediately after the server sends
+a pseudo-rectangle with the VMware Display Mode Change pseudo-encoding.
+
+The *width*  and *height* of the pseudo-rectangle specify the new width
+and height of the display. The rest of the format is described below:
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+1               ``U8``               *bits-per-sample*
+1               ``U8``               *depth*
+1               ``U8``               *color*
+1               ``U8``               *true-color*
+2               ``U16``              *max-red*
+2               ``U16``              *max-green*
+2               ``U16``              *max-blue*
+1               ``U8``               *red-shift*
+1               ``U8``               *green-shift*
+1               ``U8``               *blue-shift*
+3                                    padding
+=============== ==================== =======================
+
+VMware Virtual Machine State Pseudo-encoding
+--------------------------------------------
+
+The server sends a pseudo-rectangle with the VMware Virtual Machine
+State pseudo-encoding to notify the client of changes in the Virtual
+Machine state.
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+2               ``U16``              *vm-state-flags*
+=============== ==================== =======================
+
+*vm-state-flags* is a bitmask field. These flags describe the virtual
+machine state.
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+0               A end-user sitting at a local virtual machine console has entered fullscreen mode.
+1               A end-user sitting at a local virtual machine console has temporarily disabled VNC updates.
+=============== =======================================================
 
 Extended Clipboard Pseudo-Encoding
 ----------------------------------


rfbproto.rst
```diff
@@ -2343,6 +2343,13 @@ Number       Name
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
 -763 to -768 `JPEG Subsampling Level Pseudo-encoding`_
 0xc0a1e5ce   `Extended Clipboard Pseudo-encoding`_
```
CendioOssman: Please try to keep this table sorted. So this entry should be below the new ones.

rfbproto.rst
```diff
@@ -2343,6 +2343,13 @@ Number       Name
...
+0x574d5664   `VMWare Cursor Data Pseudo-encoding`_ (``VMWd``)
```
CendioOssman: I don't think we need the text version of the encoding numbers. Or do they add something useful I'm overlooking?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
     4               ``U32``                         *size*
     *size*          ``U8`` array                    *data*
     =============== ==================== ========== ===================
+
+VMWare Cursor Data Pseudo-encoding
+----------------------------------
+
+Sends the cursor bits and mask where cursor hot spot is defined by x,y position of image.
```
CendioOssman: I guess you mean pseudo-rectangle and not image? And if you can use the keywords from earlier definitions. In this case it would be `*x-position*` and `*y-position*`.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+*cursor-type* is either 0 to indicate color cursor, or 1 to indicate an alpha cursor.
+
+A color cursor type is followed by:
+
+======================================= =================== =============
+No. of bytes                            Type                Description
+======================================= =================== =============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor bits*
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor mask*
+======================================= =================== =============
```
CendioOssman: How are the bits and mask defined?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+
+An alpha cursor type is followed by:
+
+======================================= =================== ==============
+No. of bytes                            Type                Description
+======================================= =================== ==============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor bytes*
+======================================= =================== ==============
+
+Alpha cursors should be drawn by compositing the cursor image into the framebuffer.
```
CendioOssman: Is this necessary? Can't the client decide on how to render it?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+
+VMWare Cursor State Psuedo-encoding
```
CendioOssman: "pseudo"

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+=============== ==================== ========== =======================
+
+*cursor-state* a bit-field which describes the state of the cursor:
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+`0x01`          Cursor visible.
+`0x02`          Cursor absolute.
+`0x04`          Cursor warp.
```
CendioOssman: I assume this is actually bit 3 (which has value 0x04)?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+Bit             Description
+=============== =======================================================
+`0x01`          Cursor visible.
+`0x02`          Cursor absolute.
+`0x04`          Cursor warp. 
+=============== =======================================================
+
+If cursor visible is not set, the cursor should not be rendered into the framebuffer until another cursor state message is received that turns the cursor back on.
+
+The cursor absolute bit indicates whether the virtual machine is using an absolute or relative mouse.
```
CendioOssman: What does that mean to a client though? What should it do differently?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+
+The cursor warp bit is set when the virtual machine artificially moves the position of the cursor.
```
CendioOssman: Same here. What should a client do when it sees this?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+`0x01`          Cursor visible.
+`0x02`          Cursor absolute.
+`0x04`          Cursor warp. 
+=============== =======================================================
+
+If cursor visible is not set, the cursor should not be rendered into the framebuffer until another cursor state message is received that turns the cursor back on.
```
CendioOssman: Lines should be wrapped at column 72.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+The cursor absolute bit indicates whether the virtual machine is using an absolute or relative mouse.
+
+The cursor warp bit is set when the virtual machine artificially moves the position of the cursor.
+
+VMWare Cursor Position Psuedo-encoding
+--------------------------------------
+
+This psuedo-encoding takes no additional data. The (x, y) coordinates of the pseudo-rectangle defines the new position of cursor hot spot (not the top left corner of cursor image).
+
+VMWare Keyboard Typematic Pseudo-encoding
```
CendioOssman: Is there some other references calling this typematic? Otherwise I would suggest using the more common key repeat.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+-----------------------------------------
+
+This psuedo-encoding describes the typematic info for the virtual keyboard device.
```
CendioOssman: Although this originates with VMware and they use it for virtual stuff, others might not. So avoid referencing virtual devices or machines in the descriptions.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+
+=============== ==================== ========== =======================
+No. of bytes    Type                 [Value]    Description
```
CendioOssman: If you don't use `[Value]` then leave it out.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+=============== ==================== ========== =======================
+
+*typematic-on* is 1 if the VNC client should handle key repeat, 0 if not.
+*period*  defines the period to wait between key repeats.
+*delay* defines the delay for the first key repeat.
+
+VMWare Keyboard LED State Pseudo-encoding
+-----------------------------------------
+
+This psuedo-encoding describes the keyboard LED state of the virtual machine.
```
CendioOssman: I'm lacking some semantics here. Please look at the existing LED state extension for an example. Then check how this matches VMware's documented or observed behaviour.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+`0x01`          Scroll Lock
+`0x02`          Num Lock
+`0x04`          Caps Lock
+=============== =======================================================
+
+VMWare Display Mode Change Pseudo-encoding
+------------------------------------------
+
+The server changes the desktop size by sending a pseudo-rectangle with the Display Mode Change pseudo-encoding.
+This encoding is used for scenarios not addressed by the DesktopSize pseudo-encoding.
```
CendioOssman: A link here would be nice.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+Specifically, the DesktopSize encoding does not allow the VNC server to redefine both the color depth and size of the framebuffer.
```
CendioOssman: In other words any update following this will be using the newly specified pixel format?

CendioOssman: Or perhaps even the next rectangle in the same update?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+2               ``U16``                         *max-red*
+2               ``U16``                         *max-green*
+2               ``U16``                         *max-blue*
+1               ``U8``                          *red-shift*
+1               ``U8``                          *green-shift*
+1               ``U8``                          *blue-shift*
+3                                               padding
+=============== ==================== ========== =======================
+
+This block occurs at the beginning of the file, before every keyframe, and whenever the format of the stream changes (for example, when the virtual machine changes display resolution).
```
CendioOssman: This is a protocol specification and we should not be referencing any stream dumps. Such use should be kept separate from the main document.

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+VMWare Virtual Machine Errata Pseudo-encoding
+---------------------------------------------
+
+=============== ==================== ========== =======================
+No. of bytes    Type                 [Value]    Description
+=============== ==================== ========== =======================
+2               ``U16``                         *vm-errata-flags*
+=============== ==================== ========== =======================
+
+*vm-errata-flags* is a bitmask field. These flags describe various pieces of virtual machine state. They are useful in the implementation of a remote VNC client.
```
CendioOssman: "Errata" seems incorrect given that description. "Server State" perhaps?

rfbproto.rst
```diff
@@ -3756,3 +3763,151 @@ be set.
...
+
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+`0x01`          A end-user sitting at a local virtual machine console has entered fullscreen mode.
```
CendioOssman: What does this mean for the client?

rfbproto.rst
qmfrederik: In this case, it's a FourCC code, and often FourCC codes are referred to by their text value instead of the numeric value. That said, it's a convention used by media files which may not carry over to the RFB protocol specification.

I often use the text value when reading stream dumps, I can remove them if you want.

rfbproto.rst
qmfrederik: OK - I removed this wording.

rfbproto.rst
qmfrederik: Fixed.

rfbproto.rst
qmfrederik: You're right, fixed.

rfbproto.rst
qmfrederik: It's mainly information the server sends to the client - i.e. the client should not do anything different, so I added wording to that effect.

rfbproto.rst
qmfrederik: OK, changed to key repeat.

rfbproto.rst
qmfrederik: Fixed.

rfbproto.rst
qmfrederik: OK, changed throughout.

rfbproto.rst
qmfrederik: Added

rfbproto.rst
qmfrederik: Yes, the change is immediate. I added wording to clarify.

rfbproto.rst
qmfrederik: Sure, I removed this paragraph.

rfbproto.rst
qmfrederik: Yup, make sense.

rfbproto.rst
qmfrederik: As far as I know, it's for information purposes only. VMware clients may render a banner at the top of the screen to display this information to the end user.

qmfrederik:
Thanks for taking the time to review this and provide feedback; I updated the PR accordingly.

rfbproto.rst
CendioOssman: I'd say drop them. The protocol isn't really designed around FourCC codes, so it will generally not be helpful.

rfbproto.rst
```diff
@@ -2342,6 +2342,13 @@ Number       Name
 -314         `Cursor With Alpha Pseudo-encoding`_
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
 -763 to -768 `JPEG Subsampling Level Pseudo-encoding`_
+0x574d5664   `VMWare Cursor Pseudo-encoding`_ (``VMWd``)
+0x574d5665   `VMWare Cursor State Pseudo-encoding`_ (``VMWe``)
+0x574d5666   `VMWare Cursor Position Pseudo-encoding`_ (``VMWf``)
+0x574d5667   `VMWare Key Repeat Pseudo-encoding`_ (``VMWg``)
+0x574d5668   `VMWare LED state Pseudo-encoding`_ (``VMWh``)
+0x574d5669   `VMWare Display Mode Change Pseudo-encoding`_ (``VMWi``)
+0x574d566a   `VMWare Virtual Machine Errata State Pseudo-encoding`_ (``VMWj``)
```
CendioOssman: This link doesn't work, so there is some difference between it and the actual heading.

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+pixel is value.
+
+======================================= =================== =============
+No. of bytes                            Type                Description
+======================================= =================== =============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor bits*
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor mask*
+======================================= =================== =============
+
+The data for a color cursor consists of *width* * *height* pixel values:
```
CendioOssman: Did you mean alpha cursor here?

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+
+======================================= =================== ==============
+No. of bytes                            Type                Description
+======================================= =================== ==============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor bytes*
```
CendioOssman: Is this really true for alpha cursors? Which bits are alpha in that case?

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+
+The data for a color cursor consists of *width* * *height* pixel values
+followed by a bitmask, where each bit in the bitmask defines whether the
+pixel is value.
+
+======================================= =================== =============
+No. of bytes                            Type                Description
+======================================= =================== =============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor bits*
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor mask*
```
CendioOssman: Is this really true for the mask? It seems like it would use a more compact encoding.

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+======================================= =================== ==============
+No. of bytes                            Type                Description
+======================================= =================== ==============
+*width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor bytes*
+======================================= =================== ==============
+
+VMWare Cursor State Pseudo-encoding
+-----------------------------------
+
+A server sets the cursor state by sending a psuedo-rectangle with the
```
CendioOssman: Another "psuedo" :)

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+3               Cursor wrap. 
+=============== =======================================================
+
+If cursor visible is not set, the cursor should not be rendered into
+the framebuffer until another cursor state message is received that
+turns the cursor back on.
+
+The cursor absolute bit indicates whether the virtual machine is using
+an absolute or relative mouse. This value is for information purposes
+only.
```
CendioOssman: Any idea why they added this?

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+
+=============== ==================== =======================
+No. of bytes    Type                 Description
+=============== ==================== =======================
+2               ``U16``              *key-repeat-enabled*
+4               ``U32``              *period*
+4               ``U32``              *delay*
+=============== ==================== =======================
+
+*typematic-repeat-enabled* is 1 if the VNC client should handle key
```
CendioOssman: This reference needs to be updated.

rfbproto.rst
```diff
@@ -3756,3 +3763,193 @@ be set.
...
+2               ``U16``              *vm-state-flags*
+=============== ==================== =======================
+
+*vm-state-flags* is a bitmask field. These flags describe the virtual
+machine state.
+
+=============== =======================================================
+Bit             Description
+=============== =======================================================
+`0x01`          A end-user sitting at a local virtual machine console has entered fullscreen mode.
```
CendioOssman: This still has the bit table style that is easy to get wrong. :)

CendioOssman:
I just noticed that the existing bit tables start at `0`, not `1`. So please have a look at that as well. :/

CendioOssman:
Oh, and it's "VMware", not "VMWare".

CendioOssman:
Ping @qmfrederik

rfbproto.rst
qmfrederik: I've never used that feature, but here is what the [docs have to say](https://docs.vmware.com/en/VMware-Horizon-Client-for-Windows/4.8/horizon-client-windows-user/GUID-D49EF6B2-14D5-437A-AF31-F806161B34F9.html)

> In most circumstances, if you are using applications that do not require 3D rendering, Horizon Client transmits information about mouse pointer movements by using absolute coordinates. Using absolute coordinates, the client renders the mouse movements locally, which improves performance, especially if you are outside the corporate network.
>
> For work that requires using graphics-intensive applications, such as AutoCAD, or for playing 3D video games, you can improve mouse performance by enabling the relative mouse feature, which uses relative, rather than absolute, coordinates.
>
> When the relative mouse feature is enabled, performance might be slow if you are outside the corporate network, on a WAN.

rfbproto.rst
qmfrederik: [This document](https://wiki.multimedia.cx/index.php/VMware_Video) describes the following algorithm for drawing the cursor:

```
 dst[i] = (dst[i] & bits[i]) ^ mask[i];
```

So it does appear that amount of bytes in a mask is equal to the amount of bytes in a pixel.

CendioOssman: Ah. Alright. In that case it is really an AND mask and an XOR mask (the classical model for a cursor). I think it is best if we use that terminology as the other cursor types have normal pixel colour values.

rfbproto.rst
```diff
@@ -3764,11 +3764,11 @@ be set.
     *size*          ``U8`` array                    *data*
     =============== ==================== ========== ===================
 
-VMWare Cursor Pseudo-encoding
+VMware Cursor Pseudo-encoding
 -----------------------------
```
CendioOssman: All of these sections should be before the "Extended Clipboard Pseudo-encoding" section to follow the order of the initial table.

rfbproto.rst
```diff
@@ -2342,13 +2342,13 @@ Number       Name
...
-0x574d5668   `VMWare LED state Pseudo-encoding`_
-0x574d5669   `VMWare Display Mode Change Pseudo-encoding`_
-0x574d566a   `VMWare Virtual Machine Errata State Pseudo-encoding`_
+0x574d5664   `VMware Cursor Pseudo-encoding`_
+0x574d5665   `VMware Cursor State Pseudo-encoding`_
+0x574d5666   `VMware Cursor Position Pseudo-encoding`_
+0x574d5667   `VMware Key Repeat Pseudo-encoding`_
+0x574d5668   `VMware LED state Pseudo-encoding`_
+0x574d5669   `VMware Display Mode Change Pseudo-encoding`_
+0x574d566a   `VMware Virtual Machine Errata State Pseudo-encoding`_
```
CendioOssman: I'm afraid this link is still broken. I think it is just "Errata" that needs to be removed.

rfbproto.rst
```diff
@@ -3796,19 +3795,29 @@ No. of bytes                            Type                Description
 *width* * *height* * *bytesPerPixel*    ``PIXEL`` array     *cursor mask*
 ======================================= =================== =============
 
-The data for a color cursor consists of *width* * *height* pixel values:
+Color cursors can be drawn using the following code:
+
+::
+
+  dst[i] = (dst[i] & bits[i]) ^ mask[i];
```
CendioOssman: Very nice with an example. But I would suggest names that more clearly refer to the previous table. Other parts of the document have avoided spaces in the names specifically so we can easily refer to them elsewhere. So I would suggest `and-mask` and `xor-mask` and the example would then be `dst[i] = (dst[i] & and-mask[i]) ^ xor-mask[i];`

rfbproto.rst
CendioOssman: Hmmm... I wonder if this means that the coordinates in the "VMware Cursor Position Pseudo-encoding" will change...

We can leave that for someone else to figure out. :)

qmfrederik:
@CendioOssman Thanks for the thorough review! I made changes to act upon the feedback.

CendioOssman:
Nice. Thank you for doing to work to get this in a really good shape. I noticed one final thing in my final read through; it says that the alpha cursor consists of a `PIXEL` array, which I assume isn't true. It should be an `U8` or `U32` array given that each pixel has a fixed size.

After that we need to squash everything in to a single commit. Can you do that and force push to this PR? Or otherwise I can fix it and do a manual commit.

qmfrederik:
Sure thing! I changed the alpha cursor to be an `U8` array, and squashed everything into a single commit.

CendioOssman:
All done. Thank you for your contribution. :)

PR #16, opened 2018-10-17
add ultravnc single click prompt

jianye-dev:


rfbproto.rst
```diff
@@ -374,6 +374,7 @@ Number      Name
 2           `VNC Authentication`_
 16          `Tight Security Type`_
 19          `VeNCrypt`_
+107         `UltraVNC Single Click Prompt`_
```
CendioOssman: Hmmm.... I'm looking at the IANA link and this number isn't registered. If UltraVNC are using it without any other handshake then they really need to register it. Do they have other authentication numbers they use? We need to have this allocated with IANA before we can start recommending it to more implementations.

rfbproto.rst
```diff
@@ -419,6 +420,7 @@ handshaking was successful.
 No. of bytes    Type    [Value]     Description
 =============== ======= =========== ===================================
 4               ``U32``             status:
+..                      -1          restart
```
CendioOssman: The status is unsigned, so for clarity I think it is better to write this as `0xffffffff`.

CendioOssman: We don't really have an allocation policy for these. I think we need to mail IANA and see if they can set up a registry. We don't want to have conflicts here.

CendioOssman: How is this restart defined? What should a client do when it sees this result?

rfbproto.rst
```diff
@@ -783,6 +785,28 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+UltraVNC Single Click Prompt
+----------------------------
+
+UltraVNC SC Prompt authentication is to be used and protocol data is to be sent
+unencrypted. The server sends a computer information:
+
+========================== ============= ==========================
+No. of bytes               Type          Description
+========================== ============= ==========================
+4                          ``U32``       *prompt-length*
```
CendioOssman: Didn't you state that this is in little-endian in your PR to noVNC?

rfbproto.rst
```diff
@@ -783,6 +785,28 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+*prompt-length*            ``U8`` array  *prompt-string*
```
CendioOssman: And what is in this? What should the client do with it?

rfbproto.rst
```diff
@@ -783,6 +785,28 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+========================== ============= ==========================
+
+The client sends a non-zero value to accept.
```
CendioOssman: Why does the client need to accept something?

CendioOssman (changes requested):
Thanks for your contribution. There is unfortunately still not quite clear how to use this extension. Please have a look at the comments and see if you can make it more obvious for someone implementing a client or server.

rfbproto.rst
secureinfo: The server will resend some new authentication types to be selected by the client and then continue the authentication process.

rfbproto.rst
CendioOssman: Let's start with the allocations and deal with the behaviour separately. Can you mail IANA and get these numbers allocated? A quick glance suggests that UltraVNC has laid claim to the entire `0x68` to `0x7f` range.

rfbproto.rst
CendioOssman: You don't need to be a UltraVNC developer in these cases. Just tell IANA that you've found that UltraVNC uses these numbers and that they should be noted in their lists. I've done that in the past without issue. :)

CendioOssman: The question is if this is really a "restart", or a "continue". I.e. should the client throw away any information already agreed upon? Or is it a request for the server for several steps?

CendioOssman:
@secureinfo, are you still looking at getting this completed?

CendioOssman:
No response. Closing.

Neustradamus:
@RudiDeVos: Have you seen this PR and comments?

PR #22, opened 2019-10-23
Add specification for Tight Unix Login Auth

samhed:
Regarding server support, see this note from @dcommander:

> As far as I'm aware, the TurboVNC Server is the only VNC server that supports the Unix Login authentication sub-type. That feature was added to the TurboVNC Server nearly 10 years ago by developers at UCAR. Unix Login authentication was apparently implemented in TightVNC 1.3.8, but it was removed in TightVNC 1.3.9 (quentinsf/vnc2dl@bfab378) for unknown reasons. The more modern TightVNC v2.x Server for Windows does not support it either.

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -601,7 +601,7 @@ Code    Vendor      Signature       Description
 19      "``VENC``"  "``VENCRYPT``"  `VeNCrypt`_
 20      "``GTKV``"  "``SASL____``"  Simple Authentication and Security
                                     Layer (SASL)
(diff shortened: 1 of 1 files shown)


samhed:
As a side-note, I sent a request to IANA asking if we could officially assign 129 as an RFB security type.

CendioOssman:
Looks good. But the commit message needs to be adjusted to follow our desired format.

CendioOssman:
There is also a lower case "a" for "authentication" in the initial paragraph, which doesn't match the rest of the name.

samhed:
Fixed!

PR #23, opened 2019-12-20
Fix custom compression level range

samhed:
The actual range is 0-9 as described in https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#76encodings

I verified that the zlib implementation accepts 0-9 as well.


CendioOssman:
Commit message doesn't follow the desired format, but otherwise LGTM.

samhed:
Should be fixed now 👍

PR #29, opened 2020-09-01
Clarify semantics around SetDesktopSize message / ExtendedDesktopSize encoding

### LINKED ISSUE #28: Unclear description wrt support multi-head support clients
I'm trying to understand how to support dynamic client initiated desktop resize, without also supporting multi-head configuration. 

The ExtendedDesktopSize message should support this use case but some parts of the description are not entirely clear to me.

AFAICT, conceptually in multi-head mode, the server has a single huge framebuffer that is sub-divided into multiple non-overlapping regions. After cutting out the individual regions for individual heads, there may remain some unused gaps, as different sized heads can't be tiled precisely gapless.  We would benefit from describing this a bit more explicitly I think.

For a traditional client, all heads will thus be merged together into a single huge view, and they'll still get "DesktopSize" encoding FramebufferUpdate messages when multi-head config changes, and so everything just works normally.

For a client which supports ExtendedDesktopSize, the FramebufferUpdate message, the x, y co-ordinates can be used to determine which head the update corresponds to.  They will only get "ExtendedDesktopSize" encoding in FramebufferUpdate messages, never a "DesktopSize" encoding.

When gets fuzzy is the scenario where a client supports ExtendedDesktopSize but does not want to implement multi head.

In this section:

https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#extendeddesktopsize-pseudo-encoding

> Note that a simple client which does not support multi head does not need to parse the list of screens and can simply display the entire framebuffer.

The problem with this recommendation is that AFAICT, there is no way for the client to figure out the dimensions of the entire framebuffer.  Such a client won't get the "DesktopSize" encoding FramebufferUpdate messages, so the only way to determine the entire framebuffer size is indeed to parse the list of screens and calculate their cumulative dimensions.  

What am I missing here ?  It seems this quoted sentence is simply wrong and needs rewriting to tell clients to calculate the union of dimensions of each screen.

berrange:
This adds some cross linking to the "Screen Model" section which helps
readers interpret the ExtendedDesktopSize encoding and SetDesktopSize
client to server message.

Fixes #28
Signed-off-by: Daniel P. Berrangé <berrange@redhat.com>

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1515,7 +1515,9 @@ SetDesktopSize
 
 Requests a change of desktop size. This message is an extension and
 may only be sent if the client has previously received an
-*ExtendedDesktopSize* rectangle.
+*ExtendedDesktopSize* rectangle. This message also enables clients
+to handle configuration of a multi head screen layout within the
+framebuffer extents, as described in `Screen Model`_.
 
 The server must send an *ExtendedDesktopSize* rectangle for every
 *SetDesktopSize* message received. Several rectangles may be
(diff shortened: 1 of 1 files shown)


samhed (approved):
looks good

PR #30, opened 2020-09-01
Clarify that encoding -261 (LED State) is from the QEMU range

berrange:
Encodings -257 to -272 are a range assigned to the QEMU project.

This also distinguishes it more explicitly from the similar
VMWare LED state encoding.

Signed-off-by: Daniel P. Berrangé <berrange@redhat.com>

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1278,7 +1278,7 @@ used:
   where possible. Instead they should interpret each character-based
   keysym according to its case. Note that some applications directly
   examine the state of CapsLock and NumLock and an extension such as
-  `LED State Pseudo-encoding`_ or `VMware LED State Pseudo-encoding`_
+  `QEMU LED State Pseudo-encoding`_ or `VMware LED State Pseudo-encoding`_
   could be needed for correct behaviour.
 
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm a bit opposed to the first rationale as I think we should have a spirit of cooperation rather than reserving extensions for specific projects. But I agree with the second part about avoiding confusion with the similar extension developed by VMware.

CendioOssman (changes requested):
It seems there are more references to the old name as I get errors when trying to generate documents:

```
./rsttool rst2html "rfbproto.rst" "rfbproto.html"
rfbproto.rst:1277: (ERROR/3) Unknown target name: "led state pseudo-encoding".
```

berrange:
I didn't consider this naming to imply  non-co-operation, rather it is just a reflection of the way certain ranges of numbers were assigned to certain projects to define usage of. All the other encodings we've defined are named according to this same scheme. Co-operation between projects is inherently implied by virtue of us defining this collaborative spec.

berrange:
I pushed the update, but can't dismiss the "1 change request" message

CendioOssman (approved):
Sorry about that. I missed that you also made a push.

PR #31, opened 2020-09-01
Document the SASL security type

berrange:
This security type was implemented by QEMU (server side) and GTK-VNC
(client side), using Cyrus-SASL as the supporting library. These impls
are considered the reference implementations from which this protocol
specification was then written.

This was previously submitted 10 years ago on the mailing list

https://www.mail-archive.com/tigervnc-rfbproto@lists.sourceforge.net/msg00259.html

and for some reason I forgot to ever re-submit.

The original spec had a crazy flaw in it too - it documented a "continue-flag" taking values 1 (continue another step), and 0 (finished SAL). It should have had the exact opposite,  a "complete-flag" taking values 0 (continue another step), and 1 (finished SAL).

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -377,6 +377,7 @@ Number      Name
 13          `RSA-AES Two-step Security Type`_
 16          `Tight Security Type`_
 19          `VeNCrypt`_
+20          `SASL`_
 22          `xvp Authentication`_
 30          `Diffie-Hellman Authentication`_
 113         `MSLogonII Authentication`_
@@ -395,7 +396,6 @@ Number      Name
 14-15       RealVNC
 17          Ultra
 18          TLS
-20          SASL
 21          MD5 hash authentication
 23          Secure Tunnel
 24          Integrated SSH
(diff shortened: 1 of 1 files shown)


rfbproto.rst
```diff
@@ -800,11 +800,349 @@ is completed.
...
+*mechlist-length*  ``U8`` array                    *mechlist-string*
+================== =============================== =======================
+
+The *mechlist-string* is a list of SASL mechanism names, each separated by
+a comma. The *mechlist-length* is the number of characters in the string,
+not including the trailing ``\0``, which is not sent on the wire. Some
+example *mechlist-string* values are::
+
+  DIGEST-MD5,GSSAPI
+  ANONYMOUS,KERBEROS_V4,DIGEST-MD5
```
CendioOssman, line 845: I'm afraid I'm no too familiar with SASL. Are these guaranteed to be ASCII or do we have an encoding ambiguity here?

rfbproto.rst
```diff
@@ -800,11 +800,349 @@ is completed.
...
+
+The *mechlist-string* is typically generated on the server by a call
+to the SASL library function::
+
+  err = sasl_listmech(vs->saslconn,
+                      NULL,
+                      "",  /* Prefix */
+                     ",", /* Separator */
+                     "",  /* Suffix */
+                     &mechlist, NULL, NULL);
```
CendioOssman, line 855: Somethings off with the indentation here.

rfbproto.rst
```diff
@@ -800,11 +800,349 @@ is completed.
...
+*mechname-length*   ``U8`` array                    *mechname-string*
+4                   ``U32``                         *clientout-length*
+*clientout-length*  ``U8`` array                    *clientout-data*
+=================== =============================== =======================
+
+The distinction between NULL and "" (empty string) in the *clientout-data*
+is critical to preserve. Thus the *clientout-length* will specify the length
+of the *clientout-data* INCLUDING the trailing '\0', such that the empty
+string has *clientout-length* equal to 1, while NULL data will be transmitted
+with *clientout-length* equal to 0.
```
CendioOssman, line 900: It sounds like these things should be treated as opaque blobs, not strings. So perhaps avoid talking about terminating `\0`s?

rfbproto.rst
```diff
@@ -800,11 +800,349 @@ is completed.
...
+
+When sending the server reply the following logic is compliant with the
+wire protocol described above::
+
+      write_u32(strlen(mech))
+      write_u8(mech, strlen(mech))
+
+      if (clientout != NULL) {
+         write_u32(strlen(clientout) + 1)
+         write_u8(clientout, strlen(clientout) + 1)
```
CendioOssman, line 926: This looks wrong since it seems to do the mistake of seeing everything as strings. `sasl_client_start()` gives you an exact length of the data. Why not use that instead of trying to figure it out yourself?

rfbproto.rst
```diff
@@ -800,11 +800,349 @@ is completed.
...
+      else
+         write_u8(0)
+    } else {
+      write_u8(0)
+      write_u8(1)
+    }
+
+If *complete-flag* is 1, then the server continues with the
+`SASL server result check`_ phase, otherwise it waits for a further
+`SASL client step message`_.
```
CendioOssman, line 1105: There's a lot of duplication here. Can we merge some of this together better?

CendioOssman (changes requested):
Looks mostly good, thanks. But I have some concerns that I've noted.

rfbproto.rst
berrange: Mechanism names are standardized in various RFCs. I believe they've only ever used ASCII for them and can't imagine a reason why they'd need to use 8-bit chars of any kind

rfbproto.rst
berrange: There is some complication here because I'm trying to define the behaviour that matches the actual implementation done by QEMU and GTK-VNC. They probably should have treated things as opaque, but in practice they're sending an extra tailing \0 in most cases. This example code doesn't entirely match what they do though, so I'll try to redo this again.

CendioOssman: I can understand wanting to match existing implementations, but SASL is an already established thing outside of QEMU/GTK-VNC. So if that implementation is incompatible with other ones, then we really should think about a strategy to get that fixed. Having "I can't believe it's not real SASL" is just asking for issues, given how much GSS abstracts things away.

rfbproto.rst
berrange: The RFC limits them to A-Z, 0-9, - & _

rfbproto.rst
berrange: Yes, we need to treat them as opaque binary blobs. What I was trying (and failing) to correctly describe is that when "clientout-data" is non-NULL, we pad the length & data sent on the wire with an extra NUL byte. This was based on the belief that when de-serializing we want to preserve the distinction between ``clientout-data==NULL + clientout-len==0`` and ``clientout-data==non-NULL + clientout-len==0``.

For mechanisms where the data is purely binary there is no such distinction, as `clientout-data==non-NULL + clientout-len==0`` is conceptually invalid. For mechanisms where the data is actually a string, `clientout-data==non-NULL + clientout-len==0`` could indicate an empty string, which is technically distinct from a NUL string. 

It is unclear in retrospect whether this actually matters to mechanisms in reality, but this is the way I designed & implemented the SASL security type spec, and so we need to keep this extra NUL padding otherwise we break compatibility, and would have to request a brand new security type allocation.

I've re-written the text significantly to try to clarify this better.

rfbproto.rst
berrange: I've cut out some of the duplicated text, by referring back to earlier paragraphs.

berrange:
> @berrange, still interested in getting this finished? :)

I've pushed an update of this spec addition

rfbproto.rst
CendioOssman: Looks good!

rfbproto.rst
CendioOssman: I guess this pseudo-code needs to be updated as well.

Or perhaps just omit it? I think the description is clear enough without it.

CendioOssman: Same thing with the other pseudo-code, I guess.

rfbproto.rst
```diff
@@ -812,11 +812,352 @@ handshake is completed.
...
+with the server. The encryption parameters are not satisfactory, it must
+drop the connection.
+
+If data encryption is enabled, and suitable for the client, all future
+messages transmitted over the RFB protocol must be passed through the
+``sasl_encode`` API, and all messages received from the server AFTER the
+forthcoming `SecurityResult`_ message must be passed through ``sasl_decode``.
+
+The client now proceeds to wait for the ``SecurityResult`` message, to
+determine whether the server considers the negotiation successful.
```
CendioOssman, line 1136: What happens if the client is happy, but the server is not? Is the `SecurityResult` message encrypted or not?

CendioOssman: Never mind, I re-read the previous paragraph better. :)

CendioOssman:
I think this is looking really good now! Can we just get that pseudo-code looked at, and we should be ready to merge.

rfbproto.rst
CendioOssman: I think so, yes. If nothing else, information in two places carries the risk of being out of sync. At which point it becomes less clear instead of more. :)

berrange:
Re-pushed with redundant code snippets removed.

CendioOssman (approved):
Great! Thanks for getting this finished!

PR #32, opened 2020-12-08
Add ExtendedDesktopSize error code for qemu.

kraxel:
Signed-off-by: Gerd Hoffmann <kraxel@redhat.com>

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -3533,11 +3533,23 @@ Code    Description
 1       Resize is administratively prohibited
 2       Out of resources
 3       Invalid screen layout
+4       Request forwarded (might complete asyncronously)
 ======= ===============================================================
 
 This field shall be set to zero by the server and ignored by clients
 when not defined. Other error codes may be added in the future and
 clients must treat them as an unknown failure.
+
+The "Request forwarded" error code is used when the server is not in
+direct control of the screen layout and is unable to determine if the
+request will succeed or not.  An example for this situation is a
+virtual machine monitor like qemu which can only forward the request
+to the guest, but it is in the hands of the guest to actually respond
+to the request.  In case the request succeeds the server will send
+another *ExtendedDesktopSize* message.  Note that there can be a
+longer delay and even continuous screen updates before the request
+succeeds, for example in case the request comes in while the guest is
+booting.
 
 The *width* and *height* indicates the new width and height of the
 framebuffer.


CendioOssman:
Thanks. This addition looks good. It would be nice if we can avoid describing this strictly in terms of QEMU though. Could we rephrase things to something that would also be a good description for other implementations?

I assume that if it succeeds it will at some later point be followed by a resize from the server with `x-position` set to `0`?

Have you looked at how existing clients will behave when seeing this code? Any gotchas to be aware of?

berrange:
Yes, should expect to see an update with x-pos==0 if it later succeeds. Existing clients should treat the new as an unknown failure, but none the less honour any later server initiated resize.

The intent behind this extra code was to allow the clients to distinguish between resizes that definitely won't ever work, and resizes that will hopefully work but asynchronously. In the former case, the client can stop allowing window resizes, or switch to scaling the display, while in the latter case the client can allow resizes and not attempt to apply scaling with the expectation that the window will (likely) soon resize.

kraxel:
> Thanks. This addition looks good. It would be nice if we can avoid describing this strictly in terms of QEMU though. Could we rephrase things to something that would also be a good description for other implementations?

Well, I've tried to create a generic name to cover that.  I find it hard to describe the purpose in a useful way without that concrete virtual machine example though.
 
> I assume that if it succeeds it will at some later point be followed by a resize from the server with `x-position` set to `0`?

Yes.  Usually rather fast.  But in case the request comes in while the virtual machine boots and the display driver is not yet loaded it may take several seconds and a few screens of boot messages scrolling by before the switch actually happens.

> Have you looked at how existing clients will behave when seeing this code? Any gotchas to be aware of?

tivervnc logs the error code to the terminal (like any other error code) and seems to not care much otherwise.
didn't test other clients so far.

kraxel:
For reference: the qemu patch (v2).
https://patchwork.ozlabs.org/project/qemu-devel/patch/20201208115737.18581-9-kraxel@redhat.com/

CendioOssman:
> > Thanks. This addition looks good. It would be nice if we can avoid describing this strictly in terms of QEMU though. Could we rephrase things to something that would also be a good description for other implementations?
> 
> Well, I've tried to create a generic name to cover that. I find it hard to describe the purpose in a useful way without that concrete virtual machine example though.
> 

I think the name is fine. It's the description that gave me some cause for concern. An example of something more general would be `The "Request forwarded" error code is used when the server is not in direct control of the screen layout and is unable to determine if the request will succeed or not. E.g. when ...`.

Readers of this specification should be people focused on how the protocol behaves. Internal details about QEMU is mostly a distraction, unless they help clarify the behaviour or situation.

On the same theme, the term `screen layout` is used for this extension rather than `display resolution`.

> > I assume that if it succeeds it will at some later point be followed by a resize from the server with `x-position` set to `0`?
> 
> Yes. Usually rather fast. But in case the request comes in while the virtual machine boots and the display driver is not yet loaded it may take several seconds and a few screens of boot messages scrolling by before the switch actually happens.
> 

I think that would be useful if explicitly stated in the specification. So that other servers know how to implement this, and clients know what to expect.

rfbproto.rst
```diff
@@ -3533,12 +3533,24 @@ Code    Description
...
 clients must treat them as an unknown failure.
 
+The "Request forwarded" error code is used when the server is not in
+direct control of the screen layout and is unable to determine if the
+request will succeed or not.  An example for this situation is a
+virtual machine monitor like qemu which can only forward the request
+to the guest, but it is in the hands of the guest to actually respond
+to the request.  In case the request succeeds the server will send
+another *ExtendedDesktopSize* message.  Note that there can be a
+longer delay and even contignous screen updates before the request
```
CendioOssman, line 3550: I guess this should be `continuous`?

CendioOssman (changes requested):
Looks good. :+1:

A small typo seems to have crept in though. Could you have a look? After that I can go ahead and merge it.

CendioOssman:
Nice, thanks.

CendioOssman (approved):

PR #35, opened 2020-12-10
Record encoding/message IDs in headers

berrange:
Providing the encoding or message ID in the header for each section
makes it easy to locate parts of the spec from the table of contents.

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -446,8 +446,10 @@ Version 3.3 and 3.7
 Security Types
 ++++++++++++++
 
-None
-----
+.. _none:
+
+None (1)
+--------
 
 No authentication is needed and protocol data is to be sent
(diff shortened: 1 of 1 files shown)


CendioOssman:
This breaks all references to those sections, so I think the cost is a bit too high. :/

How useful is this though? I would assume most code would try to define constants and rarely care exactly what value things might have?

berrange:
Urgh, that's annoying about the hyperlinks :+1: 

This is more about aiding debugging - debug logs from servers/clients just mention the ID values, and its useful to be able to quickly see what they match.

CendioOssman:
Perhaps we can add some convenient links directly to the tables with the values?

berrange:
Closing since it isn't important enough problem to spend more time on

PR #39, opened 2021-02-04
Open H.264 Encoding

mdevaev:
The story:
- https://github.com/TigerVNC/tigervnc/issues/1187

The reference implementation:
- https://github.com/TigerVNC/tigervnc/pull/1194

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -2362,6 +2362,7 @@ Number       Name
 7            `Tight Encoding`_
 8            `zlibhex Encoding`_
 16           `ZRLE Encoding`_
+50           `Open H.264 Encoding`_
 -260         `Tight PNG Encoding`_
 ============ ==========================================================
 
@@ -3112,6 +3113,82 @@ are:
     =================== =============== ======= =======================
 
     Where *r* is floor((*runLength* - 1) / 255).
+
+Open H.264 Encoding
+-------------------
+
+The Open H.264 Encoding implements the widely used H.264 format
+for efficient rectangle compression for transmission over poor
+communication channels.
+
+The message looks like this:
+
+    =============== ================ ==================================
+    No. of bytes    Type             Description
+    =============== ================ ==================================
+    4               ``U32``          *length* of *data*
+    4               ``U32``          *flags*
+    *length*        ``U8`` array     *data*
+    =============== ================ ==================================
+
+The *flags* is used by the server to send additional information
+to the client. The current values of the *flags* are listed below.
+Unknown flag values should be ignored by the client.
+
+    =============== ===================================================
+    Bit             Description
+    =============== ===================================================
+    0               **ResetContext**
+    1               **ResetAllContexts**
+    2-31            Currently unused
+    =============== ===================================================
+
+The *data* is one or more H.264 frames glued together in a row.
+This encoding prescribes the use of the H.264 baseline profile
+for the lowest latency.
+
+Since H.264 is a differential format, this introduces the concept
+of context for RFB frame. The context is the state of the H.264
+encoder/decoder associated with a particular rectangle. When the server
+wants to send an H.264 rectangle, the *rect* coordinates and size
+inside a *FrameBufferUpdate* will be used to find the decoder context
+in the client. If the client finds a suitable decoder that can be used
+for the specified coordinates, the client uses it (and its internal
+state) to decode and display the new frame. The client should only
+decode frames based on the internal state of the context and ignore
+changes in the output buffer that may occur when overlapping of rects.
+When new data is received, the client context will redraw its rectangle
+and update the previously overlapped data.
+
+The server can use the *flags* to control the contexts of the client.
+All currently existing flags must be applied before decoding.
+The **ResetContext** flag requires the client to delete the current
+context with coordinates specified in *FrameBufferUpdate* and create
+a new one. The **ResetAllContexts** flag deletes all client contexts.
+
+The server can send an empty *data* and *length* equal to zero,
+in which case the client must interpret this *FramebufferUpdate*
+as a control message and apply the *flags*.
+
+The server can send multiple frames in a single *data*, which must be
+parsed as a regular H.264 stream. The server is unable to split frames
+into multiple packets. The *data* field must always contain 0 or more
+whole frames. The client must first decode all the received frames
+in a single message, and then display the result.
+
+The server must start sending *data* for the new context from I-frame
+to provide correct decoding for the client side.
+
+The client can handle decoding errors at its discretion. However,
+the recommended processing method is to ignore and wait for the correct
+frame from the server side.
+
+The client must support 64 simultaneously contexts. The server must use
+a minimum number of contexts to ensure that a single screen works.
+Ideally, one context per screen. When trying to create a new context
+beyond 64, the client must destroy the oldest context, then create
+the new one. The oldest context is the one that has not received frame
+updates for the longest time.
 
 Tight PNG Encoding
 -------------------------


CendioOssman:
Great. I'll try to have a look soon. 

Is there also code ready in Pi-KVM that can be referenced?

mdevaev:
Of course.

Here the RFB implementation: https://github.com/pikvm/kvmd/blob/master/kvmd/apps/vnc/rfb/__init__.py#L171

VNC server: https://github.com/pikvm/kvmd/blob/master/kvmd/apps/vnc/server.py

H264 compressor: https://github.com/pikvm/ustreamer/tree/master/src/ustreamer/h264

The server and the encoder work asynchronously. The H.264 stream is generated by a single encoder, and multiple VNC clients can receive this stream. Simpler thirdparty server implementations with synchronous encoding will also be able to implement this protocol just as easily. They don't even have to do frame gluing.

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+
+The *data* is one or more H.264 frames glued together in a row.
+This encoding prescribes the use of the H.264 baseline profile
+for the lowest latency.
+
+Since H.264 is a differential format, this introduces the concept
+of context for RFB frame. The context is the state of the H.264
+encoder/decoder associated with a particular rectangle. When the server
+wants to send an H.264 data, the *FrameBufferUpdate* coordinates and
+size will be used to find the decoder context in the client.
```
CendioOssman, line 3151: It's each rect that has coordinates.

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+    =============== ===================================================
+
+The *data* is one or more H.264 frames glued together in a row.
+This encoding prescribes the use of the H.264 baseline profile
+for the lowest latency.
+
+Since H.264 is a differential format, this introduces the concept
+of context for RFB frame. The context is the state of the H.264
+encoder/decoder associated with a particular rectangle. When the server
+wants to send an H.264 data, the *FrameBufferUpdate* coordinates and
```
CendioOssman, line 3150: "data" looks weird here. Did you mean to write "rectangle"?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+size will be used to find the decoder context in the client.
+If the client finds a suitable decoder that can be used for the
+specified coordinates, the client uses it (and its internal state)
+to decode and display the new frame.
+
+The server can use the *flags* to control the contexts of the client.
+The **ResetContext** flag requires the client to delete the current
+context with coordinates specified in *FrameBufferUpdate* and create
+a new one. The **ResetAllContexts** flag deletes all client contexts.
```
CendioOssman, line 3159: Before decoding the data in the rect? Or after?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+
+The server can send an empty *data* and *length* equal to zero,
+in which case the client must interpret this *FramebufferUpdate*
+as a control message and apply the *flags*.
```
CendioOssman, line 3163: I assume this is to signal the end of a stream?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+
+The server can send multiple frames in a single *data*, which must be
+parsed as a regular H.264 stream.
```
CendioOssman, line 3166: Can the server send partial frames?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+
+The server can start sending *data* for the new context from P-frames.
+In this case, the client should ignore them until it gets the first
+I-frame. However, the client should not ignore the *flags*.
```
CendioOssman, line 3170: Do we need to start getting in to details of frame types? Is this really a requirement on the server that the first rectangle for a new context should contain enough data to render every pixel within the rectangle?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+
+The client must first decode all the received frames in a single
+message, and then display the result.
```
CendioOssman, line 3173: Might be clearer to merge this with the section stating that there could be multiple frames. I assume the general meaning is that a rectangle can contain lots and lots of frames, but only the final result should be displayed?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+for the lowest latency.
+
+Since H.264 is a differential format, this introduces the concept
+of context for RFB frame. The context is the state of the H.264
+encoder/decoder associated with a particular rectangle. When the server
+wants to send an H.264 data, the *FrameBufferUpdate* coordinates and
+size will be used to find the decoder context in the client.
+If the client finds a suitable decoder that can be used for the
+specified coordinates, the client uses it (and its internal state)
+to decode and display the new frame.
```
CendioOssman, line 3154: Given the state, how does this interact with other encodings? Are they forbidden from overlapping? I would assume so. But what happens if they overlap? Is that a protocol violation? Or an implicit recent context?

CendioOssman (changes requested):
Looks very nice. But I have a few things that would be nice to get clarified.

rfbproto.rst
berrange: We should always explicitly specific the behaviour when the client's Frame Buffer Update Request had  incremental==0. In such a case, I think the server should only send complete frames and not rely on any existing client side state.

CendioOssman: So specify that it forces the server to do a `ResetAllContexts`?

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+I-frame. However, the client should not ignore the *flags*.
+
+The client must first decode all the received frames in a single
+message, and then display the result.
+
+The maximum number of simultaneously supported contexts is 255.
+When trying to create a new context beyond this value, the client
+must destroy the oldest context, then create a new one.
+The oldest context is the one that has not received frame updates
+for the longest time.
```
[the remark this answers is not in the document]
berrange: I wonder what the client can do if it needs to discard one or more of its cached contexts. As written, it can never do this as it can only discard contexts in response to the server flags.

If we declare that a frame buffer update request with incremental==0, always results in complete frames being sent by the server, then we could also use this as a way to reset contexts. eg If server gets a request with incremental==0, could we say that it *must* use  ResetAllContexts in the flags for the subsequent rects it sends ?  Or can we define a separate flag "PermittedContextReset" as a way to indicate to the client that it is permitted to reset its contexts, because the rects are all full.  This would give the client the choice to rebuild - it can skip that choice if it doesn't need to rebuild contexts and save the overhead of doing so

rfbproto.rst
berrange: That's one option - I only wonder if ResetAllContexts might be too big a hammer if the act of creating a context is a heavy weight operation ? I dont know enough to say for sure either way.

rfbproto.rst
```diff
@@ -2362,6 +2362,7 @@ Number       Name
 7            `Tight Encoding`_
 8            `zlibhex Encoding`_
 16           `ZRLE Encoding`_
+50           `Pi-KVM H.264 Encoding`_
```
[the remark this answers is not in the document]
mdevaev: No. 50 is a free number and is not used by anyone. How can we even register this?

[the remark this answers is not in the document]
mdevaev: I don't mind. But you need some name that will be different from the H264 already registered.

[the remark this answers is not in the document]
CendioOssman: There's a form you can fill out with IANA and they'll add your entry to their official list.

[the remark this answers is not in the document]
CendioOssman: Given that the other one is secret I'm more inclined to change theirs to "RealVNC H.264". First one to write a spec gets to pick the best name. :)

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+The *flags* is used by the server to send additional information to the
+client. The current values of the *flags* are listed below:
+
+    =============== ===================================================
+    Bit             Description
+    =============== ===================================================
+    0               **ResetContext**
+    1               **ResetAllContexts**
+    2-31            Currently unused
+    =============== ===================================================
```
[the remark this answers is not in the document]
mdevaev: Unknown flag values should be ignored by the client.

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+
+The *data* is one or more H.264 frames glued together in a row.
+This encoding prescribes the use of the H.264 baseline profile
+for the lowest latency.
```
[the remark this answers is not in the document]
mdevaev: Baseline is the simplest format that all encoders and decoders support, but the rest is dependent on the library used. For example, you might want to implement this in noVNC using Broadway JS decoder, but it only supports baseline.

I think that it makes sense to define a number of additional pseudo-encodings, which will mean that not only baseline can be transmitted over this protocol. The client will transmit them and the server will decide what to do with it.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: In fact, you are highly likely not to need more than a couple. H264 works efficiently for compressing large rectangles (uses its own differential encoding) and so you don't need the old RFB logic for separating them. That is, at the moment, one rectangle is equal to the size of the transmitted screen.

rfbproto.rst
mdevaev: What do you mean?

rfbproto.rst
mdevaev: Okay

rfbproto.rst
mdevaev: Before, fixed.

rfbproto.rst
mdevaev: No, the transmission can continue. This is just an opportunity for the server to send control flags.

rfbproto.rst
mdevaev: Plz explain, I don't understand what you mean)

rfbproto.rst
mdevaev: This is done so that you can have a separate encoder process and a separate server process. Since H264 encoding is a difficult task, it is effective for a single screen on the server to use a common encoder for all clients. The client can connect at any time and sending data will not start from the I-frame. In fact, this is quite normal for streaming differential video, all existing video players are required to handle such a stream.

rfbproto.rst
mdevaev: Exactly.

rfbproto.rst
mdevaev: I believe that they can overlap and there is nothing wrong with that. Here is a special case: my server implementation sends a normal stream in H264 and JPEG if it wants to display some service text on the screen.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: I suppose we should first refine our specification here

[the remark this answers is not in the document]
mdevaev: It is unlikely that the IANA will agree with this, since the person who proposed H264 there is engaged in protocol maintenance

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
 
     Where *r* is floor((*runLength* - 1) / 255).
 
+Pi-KVM H.264 Encoding
+-------------------------
```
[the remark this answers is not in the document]
mdevaev: Fixed

rfbproto.rst
```diff
@@ -3113,6 +3114,70 @@ are:
...
+The Pi-KVM H.264 Encoding is an efficient way to compress image stream
+for transmission over poor communication channels.
```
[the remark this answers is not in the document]
mdevaev: Fixed

rfbproto.rst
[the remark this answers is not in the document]
berrange: We can't merge this spec without the IANA assignment, and IIUC, IANA doesn't require you to have a spec written.

IOW There's no need to wait, just request the IANA assignment in parallel.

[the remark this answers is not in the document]
mdevaev: Okay, let's deal on the name. `Open H.264 Encoding`? Somehow it sounds strange, ambiguous.

In fact, if you use our suggested name, it would be in accordance with the already existing protocols: Tight, RealVNC and other. Why are we worse?)

rfbproto.rst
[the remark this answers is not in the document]
xornet-sl: Actually I think that the number of cached contexts can be defined by implementetion. You really don't need much and we can define some minimal number here, say 10. Implementation can have more, but it must be able to maintain at least 10 (for example). Contexts aren't heavy. In fact it is just a rect buffer + some meta

rfbproto.rst
xornet-sl: > Can the server send partial frames?

No, the server can encode and send only complete frames. One or more then one, but complete. A client should just send all these frames to H.264 decoder one by one and render the last frame.

rfbproto.rst
mdevaev: ResetAllContexts  is not a heavy operation and is not even used in my server implementation. We just thought it would be nice to have this flag for third-party implementations. It may be easier for some other server to delete all the contexts at once if it wants to reinitialize for some reason, rather than sending separate messages for each context.

xornet-sl: Client even can don't mind of frame types and send all the packets to decoder. In such case there will be just glitches on the screen until fisrt I-frame arrived. It is just ok, becouse I-frames come rather often

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: QEMU also... I use some QEMU stuff in my server, even though they were clearly not designed for this.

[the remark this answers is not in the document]
CendioOssman: And those have been a source of confusion, so we'd rather not repeat that mistake with new things. :)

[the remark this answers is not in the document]
CendioOssman: A good opportunity to poke them about the fact that they keep stuff secret. Doesn't clash well with the principle of IANA and IETF. :)

He's not really in a position to reject things. If he tries putting his employer ahead of his role at IANA then we'll have to escalate things. Not that I think he will.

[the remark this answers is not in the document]
CendioOssman: If we're sticking to just the baseline profile, then that could be part of the name to clarify things. I.e. "Baseline H.264 Encoding".

rfbproto.rst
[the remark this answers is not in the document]
CendioOssman: Fair enough. I assume you've tested baseline and found it sufficient for your use case?

rfbproto.rst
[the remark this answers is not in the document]
CendioOssman: If the number varies then there needs to be some way of the client telling the server how many it supports. Which can get complex. So ideally we find the smallest number that is sufficient and stick with that.

rfbproto.rst
CendioOssman: "FrameBufferUpdate" is the entire update with potentially lots of rectangles. So this text should refer to rectangles and not the entire update.

rfbproto.rst
CendioOssman: I meant as in the end of that particular stream. So that the client can free up any context. And then any new rectangles would start fresh with a new context.

Or did you have a different use case in mind here?

rfbproto.rst
CendioOssman: The frames will be encoded to a stream of bytes. The question is if the server can split that stream up however it likes, or if it can only split on frame boundaries. I think such a requirement should be made explicit in the text.

rfbproto.rst
CendioOssman: That would violate the VNC protocol. Each update should contain enough to render the screen in a complete, consistent state. I don't think we should go away from that without good reason.

rfbproto.rst
CendioOssman: Wouldn't that break horribly if the next update only contains P-frames? They assume a certain existing contents, which we've now overwritten.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: Yes. It works great.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: I suggest 255 just for the future, but add a strong recommendation for the client and server to make do with a minimum number of contexts, ideally no more than one for the screen.

rfbproto.rst
mdevaev: Oh, yeah. We're talking about the same thing. Just the lack of data doesn't mean anything in itself, it's all about the flags. A frame with a zero flags and zero data does not have to do anything.

rfbproto.rst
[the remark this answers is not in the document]
CendioOssman: A client needs to be written for what's allowed, not what is common. So if we say 255 here, every client must write their code to handle 255. It doesn't matter if all servers will just use one.

So if less than 255 is sufficient, we should go with that IMHO.

rfbproto.rst
mdevaev: No, it can't. This doesn't make much sense and it greatly simplifies the implementation. Fixed.

rfbproto.rst
xornet-sl: > I meant as in the end of that particular stream. So that the client can free up any context. And then any new rectangles would start fresh with a new context.
> 
> Or did you have a different use case in mind here?

We don't know any future use cases. it is just a placeholder for them. Right now there are only two flags and in future can be more (for example to change the codec state or something else). It is just a control message. A client should just interpret the flags and do what they mean according this document.

CendioOssman: Alright. Since it is the server that is in control of this, but it is the client that wants it, it would be nice if we recommend a certain behaviour here. E.g. "The server **should** send a rectangle without any data and reset the context when it can in order to let the client free up resources." or something along those lines.

rfbproto.rst
xornet-sl: > The frames will be encoded to a stream of bytes. The question is if the server can split that stream up however it likes, or if it can only split on frame boundaries. I think such a requirement should be made explicit in the text.

the server can split packets only on the frame boundary. If not that then one screen will come in several update packets. I think it's just a mess. The server encodes a frame at once and it should send it as a solid packet

rfbproto.rst
xornet-sl: > That would violate the VNC protocol. Each update should contain enough to render the screen in a complete, consistent state. I don't think we should go away from that without good reason.

Ok then. What should client do when it connects to server with async encoder and the client receiving first P-frames (differential ones) in you opinion? I mean we can stick here that the client should show just a blank screen. Or it allowed to render starting glitches util I-frame arrive. It is not about the protocol itself. It is just about handling situation where I-frames have not yet been received.
The reason when server don't send initial I-frame is really good as described by @mdevaev. If one server handles more than one-two clients then encoding process for each of them would be a very hard task. Especially in situations where the server is a tiny embedded system.

rfbproto.rst
mdevaev: No, it works fine. Moreover, the server has several options for how it can handle this situation so that the client has the correct picture.
1. The server can pause the H264 encoding and send the JPEG. It will then resume encoding and send the P-frame. This frame is superimposed on the client context that stored the state of the H264 decoder before receiving the JPEG, and continues to store it further.
2. After JPEG, the server can generate an I-frame. The context will be fully updated.
3. The server can send one of the control flags to reset the context state. This method is used in my server because it does not control the H264 encoding process, but only reads the ready-made stream.

xornet-sl: It is no matter P-frame or I-frame will arrive in the next update. It is even unrelated to encoding. The server should send all overlapped rects in right order to render. With two overlapped JPEGs you overwrite overlapped region too, so the server must send both rects in order to render it right

rfbproto.rst
xornet-sl: > Alright. Since it is the server that is in control of this, but it is the client that wants it, it would be nice if we recommend a certain behaviour here. E.g. "The server **should** send a rectangle without any data and reset the context when it can in order to let the client free up resources." or something along those lines.

Why should we stick with it? It is just a control message inside the H264 protocol to do just *something*. What thing to do is described by flags not by the fact that this packet contains no rect following the header. If flags tell the client "reset the context", ok, client will reset it. If flags will instruct to do something esle in the future - client will do that if it supports it. Features that are supported here can be sent by pseudo-encodings at the start. Right now these flags are mandatory for all clients who will implement H.264 decoder. The next flags (if they will appear) will be described separately and will use standard pseudo-encodings negitiation protocol

rfbproto.rst
mdevaev: I will add to what has been said. Hypothetically, we can have the server monitor the H264 stream and parse it to see if the current frame is an I or a P-frame. Then the server would receive an UpdateRequest and wait for an I-frame to arrive from the encoder before replying to the client. This will add overhead for the server (it will have to parse the frames and wait) and it will have to remember one more thing about the client state. I think this is an unnecessary complication.

On the other hand, in case of H264, some individual packages for updating the frame no longer contain enough data to display on the screen, since these will only be P-frames, and everything will depend on whether the I-frame is contained in the client. So if you open the package in WireShark you won't see anything interesting there.

The most important reason is that this approach is absolutely natural for H264. The H264 stream allows for frame drops (even an I-frame can be lost) , and absolutely all decoders are able to handle the situation when they do not have an I-frame and get a P-frame. The client can easily process incomplete data and just show a black screen instead of a rectangle, while the server will have to have additional logic for this.

And now I'm not even talking about my own implementation. It doesn't cost me anything to do this since I have information about whether the frame is the key. Here I care about third-party implementations that may use less convenient encoders. They will have to parse the H264 stream, and this is not a very pleasant task, which will also be a waste of hardware resources.

I believe that the RFB here should act as a transport and not try to change the existing semantics established in H264. The encoder should behave normally, and the decoder should do the same. No magic at all.

rfbproto.rst
mdevaev: If we use pseudo-encodings to indicate that the client can work not only with baseline, then we should avoid this word in the name.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: If we want use pseudo-encodings to indicate that the client can work not only with baseline, then we should avoid this word in the name.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: How much do you offer? If we follow the logic of "one context = one screen", then we limit exactly the number of screens.

rfbproto.rst
mdevaev: Gotcha. Fixed.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: To be honest, I don't think there's any hidden intent here. It is possible that the existing numbers are simply reserved and we will be allowed to use them. We'll contact IANA and see what we can do about it.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: Added a clearer description of this

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: @CendioOssman I have a great news! IANA approves `50` for `Open H.264 Encoding`.

image

So we can continue our journey.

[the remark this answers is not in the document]
mdevaev: Now it has the official name `Open H.264 Encoding`.

CendioOssman:
Just an FYI, I'm a bit swamped right now so I need a few more days before I have time to check all the new comments here.

mdevaev:
@CendioOssman boop. I don't want to rush you, but we are already ready to deliver this to our users. I edited the text a bit, adding some clarifications and correcting it to match the style guide. In addition, the pikvm prefixes were removed from the patch in vncviewer to match the current specification.

CendioOssman:
Sorry, still trying to get to this, but it will likely be a bit longer. This is high on my list, but unfortunately not quite at the top.

rfbproto.rst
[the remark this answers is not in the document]
CendioOssman: RFB has just a single framebuffer, so from that point of view it just needs one. But if one is needed per monitor on the server then that is of course insufficient.

How many have you used in your server?

rfbproto.rst
CendioOssman: The specification should specify not only what is possible, but also recommended behaviour IMHO. This section becomes rather pointless without such a recommendation. Why would a server implementation ever write code that sends a rectangle with no data?

rfbproto.rst
CendioOssman: But RFB is not just a transport. It is a full display stream, just like H.264. So some extra work and glue is reasonable to expect when trying to get them to work together.

A non-incremental update is much like when a new client shows up. You need to get that client synced up to the stream as well (likely waiting for an I-frame). Until that point you should not start using this new encoding and should rely on other encodings to quickly get the client updating.

If this isn't required by the server then it should be very explicit in the specification that this encoding does not follow that basic principles of RFB that updates are consistent.

(I'd also say that such an exception makes me somewhat sceptical about including it in TigerVNC)

rfbproto.rst
CendioOssman: The problem is that H.264 can base the next frame on the previous. If you modify the pixels outside of the H.264 stream then the encoder and decoder needs to be made aware of those changes. Either by the decoder having a shared state with the RFB framebuffer, or by the RFB code writing back any changes to the H.264 decoder.

I think this places an unreasonable burden on the decoder, so I would recommend taking it out. I would guess hardware decoders would not be able to fulfil this requirement.

Either way the specification needs to be made explicit how overlapping rectangles are handled, given how stateful H.264 is. Either such things are forbidden (and hence a protocol violation), or the client needs to be prepared for them and needs to have a decoder where it can make sure things are in sync.

CendioOssman:
Most of the stuff looks ready to go. We just need to sort out the semantics around the initial (and non-incremental) updates, and the overlapping. Then we can merge this.

rfbproto.rst
mdevaev: From the point of view of third-party implementations, it may be easier for the server to send a message with some flags, and then send data without flags. Here we want to make life easier for other developers. In addition, other flags may be added in the future that may be useful without data. I just don't want to put restrictions in the protocol where it's not required.

rfbproto.rst
mdevaev: Either way, we must explicitly declare: or we forbid sending a P-frame without an I-frame (then what should the client do if it receives such a message? this must be explicitly documented), or we resolve it. Yes, I understand that this is a violation of the RFB principles, but it simplifies the implementation of the server (it does not have to monitor the flow), does not require additional effort when implementing the client (all decoders can do this).

I'm not defending my own implementation, but I want to give third-party implementations an easy life. This will allow you to implement fully asynchronous components. In addition, allowing such a situation makes the client more reliable.

Imagine that we are talking about a certain server (not mine), which contains a bug, because of which in some situations it can generate a P-frame before an I-frame. What should the client do? Ignore it? Or, you may disconnect? If it is disconnected, it may not be possible to remotely administer this server.

We are making an extension like this that was not previously available in RFB. We can impose restrictions to preserve the principles, or we can make the protocol unbreakable and easy to implement.

rfbproto.rst
mdevaev: To be honest, I don't quite understand what the problem is here. Let's say the client received the h264 data, decoded it, and then placed it on the screen. After that, the client received the data overlapping with this rectangle in any other format (maybe even h264) and placed it again. The first decoder is based on its internal state, not on what is drawn on the screen, so the arrival of new h264 data will simply cause a redraw of what is already on the screen. As a special case, I have it working with JPEG.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: Right now I'm using 1. This is because I use a single video capture device that captures a single screen. In the future, I'll add a few more, and we'll need 2 or more. I want this number to be reasonable and limited so that in the next few years it will be enough for everyone.

You wrote here that rfbproto not only describes, but also gives recommendations. Why don't we add that we support 255, and strongly recommend using this as "one context per screen"? I really don't want to be cursed by our descendants when in a few years there will be sort of 8k monitors, for which one hardware encoder will not be enough and you will have to split the entire screen into 8 or 16 contexts due to the limitations of the hardware chips (like some can't handle more than 1080p, or to speed up it's easier to split into several 800x600).

PS: Like 640 kilobytes should be enough for everyone.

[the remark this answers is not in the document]
mdevaev: I have a suggestion on this point. We can not introduce an arbitrary restriction in the form of a certain number, but make their number dynamic. At the same time, make the lower limit.

For example, we can say: at least 64 contexts should be supported. This is enough to support 4 8k screens, each of which will be encoded using 16 hardware encoders that support a maximum of 1080p. The client can support a greater number of contexts. If the server exceeds the number of contexts supported by the client, the client should treat this as an error.

The recommendation includes a strict recommendation to use a minimum number of contexts per screen (ideally, 1, if the hardware allows).

rfbproto.rst
CendioOssman: Okay

rfbproto.rst
mdevaev: Another thing is the coding errors. This can indeed happen for some reason, and the client should not disconnect if it has received a broken frame. It should wait until a good frame arrives. This is the rule that is adopted in decoding the H264 stream. From this point of view, we can consider the arrival of the P-frame before the I-frame as a special case of an error that should be ignored.

At the same time we can formulate the following two rules:
* The server must return the I-frame as the first data for the new context so that the client can decode something.
* The client must ignore invalid data that it cannot decode.

The combination of these two rules will be consistent with the spirit of the RFB protocol and still provide a reliable connection.

rfbproto.rst
mdevaev: Great! :)

rfbproto.rst
CendioOssman: To clarify, just P-frame(s) during most updates would be just fine. It's the initial and non-incremental updates that would require an I-frame. And generally clients disconnect if they get something that is not allowed by the protocol. That could mean the server can't be used, but that would be the case for most protocol bugs anyway.

And not guaranteeing an I-frame would make implementing a client **more** difficult if you want to implement a glitch-free experience for the user.

I understand this makes your server implementation easier and more efficient, but it does give a worse user experience IMHO.

rfbproto.rst
CendioOssman: You're assuming the H.264 decoder has a copy of the data on screen, which I would say is not always the case. That would mean duplication that is an inefficiency that many might like to get rid of.

So if that duplication is a required part of this encoding then it should be explicitly stated. If using the displayed data as the decoder state is allowed, then that should also be explicitly stated and likely means forbidding overlapping updates.

rfbproto.rst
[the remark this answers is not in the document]
CendioOssman: I'm not sure I follow. How does the server know if it can use more than 64 contexts in that case?

rfbproto.rst
CendioOssman: > The client must ignore invalid data that it cannot decode.

This is something that would be completely new in the RFB protocol. I think every other type of invalid data results in a disconnect. RFB was very much not designed with error recovery in mind.

H.264 was designed to operate over lossy transports. RFB requires a guaranteed transport. So there is no need to concern ourselves with such issues in this case here. Trying to compensate for server bugs in the protocol is a fool's errand IMO as such bugs will likely not be recoverable in 99% of the cases anyway. E.g. if the data length field is off by even one byte we'll have no choice but to disconnect as we have no markers to resync on.

mdevaev: The I-frame must be sent at a periodic interval in any stream, for example, every couple of seconds. To be honest, I'm not sure that you can make a stream only from a P-frame without significant image degradation. Either way, your server should periodically send an I-frame.

At the same time, absolutely any client library is able to handle any decoding errors. The arrival of the P-frame before the I-frame is exactly the same error, quite standard for the H264 decoder, because this situation often happens when live-streaming via UDP.

You can see this for yourself by looking at our PR in TigerVNC. There is no special logic for handling transmission errors, everything works naturally. Whichever library you choose, it will be able to handle this situation.

CendioOssman: And to re-iterate, this is not a blocker to get things merged. But it needs to be made explicit since it's a big deviation from how RFB normally operates.

rfbproto.rst
mdevaev: The context is not a copy of the displayed data. This is a decoder instance that stores calculated data based on I-frame and P-frame. You can't get rid of them because they are the internals of the H264 library or hardware decoder you are using.

We don't know in what form it is stored inside. But this does resolve the overlap, because even if you modified the image with a different rect, the context (and the data inside it) will remain intact and will be redrawn when a new H264 frame is received.

rfbproto.rst
mdevaev: I understand. Like you, I want this to be as true to the spirit of the RFB as possible. And yes, of course I want this to be accepted in TigerVNC, otherwise what's the point of it at all :) Do you think the two rules I have proposed are an acceptable compromise?

* The server must return the I-frame as the first data for the new context so that the client can decode something.
* The client must ignore invalid data that it cannot decode.

mdevaev: Sorry, I didn't notice your second comment.

You are right, an error at the protocol level can lead to the inability to process the message. But this can only happen if the server implementation fails, because we use TCP, which does not lose packets along the way.

The error of the encoder lies on another level, and it may be independent of us. It could be a hardware bug or something like that. The only difference is that error handling for H264 is completely free. In the sense that this is the case where ignoring errors will be easier for any implementation than imposing explicit constraints.

On the other hand, the RFB does not regulate error handling. I think this is an omission of the ancestors :) The abstract client may not be disabled when receiving failed frames. At the same time, if we say that H264 should ignore errors, we will make the client behavior more predictable and the implementation simpler.

mdevaev: PS: I did encounter hardware coding problems, it looks like "occasionally squares run across the screen". I think the user will not be happy if the client disconnects from the server because of such things.

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: For this case, we can introduce in the future a pseudo-encoding that will report the maximum number of contexts. If there is no pseudo-encoding, then 64 is used.

rfbproto.rst
CendioOssman: > * The server must return the I-frame as the first data for the new context so that the client can decode something.
> 

Sounds good. And we should also include that a non-incremental data also rests context. The general usage of non-incremental has been if the client needs to drop state, for whatever reasons.

>     * The client must ignore invalid data that it cannot decode.

I'd prefer to leave that out. If implementations want to try to compensate for server bugs then that's up to them. I don't think the specification should require them to do so.

> At the same time, if we say that H264 should ignore errors, we will make the client behavior more predictable and the implementation simpler.

Assuming they are using and already existing fancy H.264 decoder. If someone wants to write a new, simple one. I'm fairly sure they would prefer if error handling is optional. :)

Note that we aren't mandating a disconnect on errors. We've always left it up to the implementations what they want to do if someone isn't following the specification.

rfbproto.rst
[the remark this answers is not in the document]
CendioOssman: Sounds good. :)

rfbproto.rst
mdevaev: It suits me :) However, can we say in this case that ignoring decoding errors is the recommended, but not the required way to handle them?

rfbproto.rst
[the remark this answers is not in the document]
mdevaev: Cool, fixed.

rfbproto.rst
CendioOssman: I'm not convinced this isn't always the case. If I recall the filtering steps of H.264 were added to avoid having duplicate copies in the decoder as most devices wanted to do filtering before presenting the data.

So, if the is the model we're assuming I think it should be documented. I.e. rectangles using other encodings are allowed to overlap rectangles using the H.264 encoding. But subsequent updates using the H.264 encoding must ignore such other rectangles when decoding and rendering.

rfbproto.rst
CendioOssman: Fair enough.

rfbproto.rst
mdevaev: Okay, then I'll write like this:

The client should only decode frames based on the internal state of the context and ignore changes in the output buffer that may occur when overlapping of rects. When new data is received, the context will redraw its rectangle and update the previously overlapped data.

rfbproto.rst
mdevaev: Done. I also added a section on what flags should handle before frames to make the document more clear. This is important for ignoring errors.

rfbproto.rst
mdevaev: Added

mdevaev:
So, it seems that's all. When you accept this, I will inform the IANA to update the link to the encoding description, fix my server implementation (from the I-frame point of view), and we can take TigerVNC up close. I propose to open the champagne on this occasion :)

rfbproto.rst
CendioOssman: :+1:

CendioOssman (approved):
Great. Looks good to me. Thanks for working through the feedback. :)

CendioOssman:
@berrange, do you want to add anything more before we continue?

mdevaev:
I combined the paragraphs about the number of contexts into one to make it a little clearer.

mdevaev:
There seems to be no objection :) Right now, our patch for TigerVNC is fully compliant with this specification. The server behavior regarding the I-frame is also fixed.

rfbproto.rst
```diff
@@ -3113,6 +3114,82 @@ are:
...
+Since H.264 is a differential format, this introduces the concept
+of context for RFB frame. The context is the state of the H.264
+encoder/decoder associated with a particular rectangle. When the server
+wants to send an H.264 rectangle, the *rect* coordinates and size
+inside a *FrameBufferUpdate* will be used to find the decoder context
+in the client. If the client finds a suitable decoder that can be used
+for the specified coordinates, the client uses it (and its internal
+state) to decode and display the new frame. The client should only
+decode frames based on the internal state of the context and ignore
+changes in the output buffer that may occur when overlapping of rects.
```
any1, line 3159: > ... when overlapping of rects

The wording here is a bit strange.

mdevaev: Could you suggest something another? This is not my native language.

any1: Well, the thing is that I'm not exactly sure what this means.

I'm looking into implementing this for [Neat VNC](https://github.com/any1/neatvnc), so I was just reading the spec to see how to do it. This part is probably not important for my implementation, but I thought I'd point it out since I stumbled on it.

mdevaev: This means that some H264 decoders can use the output buffer as data storage for subsequent decoding, meaning it can be mutable. If your decoder uses a mutable buffer, then you should make sure that the data will not be modified by other decoders and encodings. If the buffer is immutable, then everything is fine.

mdevaev: BTW how did you find out about this encoding? This is a very, very new technology, but I'm glad that there will be another implementation.

any1: I'm subscribed to the rfbproto project, so I get notifications about PRs. I've been intending to create my own h264 encoding, but you seem to have done some of the heavy lifting for us here, which is nice.

mdevaev:
Could you merge this PR if no one has any objections? Moreover, the developers of Neat VNC server and bVNC client are interested in implementing this specification. I just don't want it hanging in the air, especially considering that all the controversial issues have been fixed and the patch for TigerVNC is also waiting for a review. We successfully tested it and found no problems.

CendioOssman:
No further comments have been made so we can go ahead and merge this. I'll need to you merge up the commits to something more final though. Please also note that we'd like a signed off line on the commit (see the README).

(and in case you're unfamiliar with Github's PRs you can just do a force push and Github will notice that you squashed things together)

mdevaev:
@CendioOssman I think I did it.

PR #40, opened 2021-02-22
Add documentation of the Replit Audio messages

lhchavez:
These messages are intended to allow clients to request the server to
send compressed audio (with the Opus and MP3 codecs) in push and pull
models.

The discussion is in https://groups.google.com/g/novnc/c/pQ6h2xLf3Kc

Reference implementation (client-side): https://github.com/novnc/noVNC/pull/1525
Reference implementation (server-side): https://github.com/replit/rfbproxy

Signed-off-by: Luis Héctor Chávez <luis@replit.com>

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1047,6 +1047,7 @@ Number      Name
 127         VMware
 128         Car Connectivity
 150         `EnableContinuousUpdates`_
+245         `Encoded Audio Client Message`_
 248         `ClientFence`_
 249         OLIVE Call Control
 250         `xvp Client Message`_
@@ -1403,6 +1404,104 @@ The server must ignore all incremental update requests
 long as continuous updates are active. Non-incremental update requests
 must however be honored, even if the area in such a request does not
 overlap the area specified for continuous updates.
+
+Encoded Audio Client Message
+---------------------------
+
+This message may only be sent if the client has previously received a
+``FrameBufferUpdate`` that confirms support for the intended
+message-type.  Every `Encoded Audio Client Message`_ begins with a
+standard header
+
+============= ======= ========= ======================================
+No. of bytes  Type    [Value]   Description
+============= ======= ========= ======================================
+1             ``U8``  245       *message-type*
+1             ``U8``            *submessage-type*
+2             ``U16``           *payload-length*
+============= ======= ========= ======================================
+
+This header is then followed by arbitrary data of length
+*payload-length*, and whose format is determined by the
+*submessage-type*. Possible values for *submessage-type* and their
+associated minimum versions are
+
+=============== =============== ================================================
+Submessage Type Minimum version Description
+=============== =============== ================================================
+0               0               `Encoded Audio Client Set Encoder State`_
+1               0               `Encoded Audio Client Enable Audio Streaming`_
+=============== =============== ================================================
+
+Encoded Audio Client Set Encoder State
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This submessage allows the client to request the server to start audio
+capture with the provided configuration, reconfigure the codec, and
+stop audio capture.
+
+============= ======= ========= ======================================
+No. of bytes  Type    [Value]   Description
+============= ======= ========= ======================================
+1             ``U8``  245       *message-type*
+1             ``U8``  0         *submessage-type*
+2             ``U16`` 8         *payload-length*
+2             ``U16``           *client-version*
+1             ``U8``            *enabled*
+1             ``U8``            *channels*
+2             ``U16``           *codec*
+2             ``U16``           *kbytes-per-sec*
+============= ======= ========= ======================================
+
+After invoking this operation, the client will receive a `Encoded Audio
+Server Start Encoder Message`_ with the result of the operation. Future
+`Encoded Audio Client Set Encoder State`_ messages can be sent after the
+encoder parameters have been set to disable the encoder, or to
+reconfigure the encoder. The *client-version* MUST be at most the maximum
+version that the server supports.
+
+Valid values for the *enabled* field are 0, which disables/stops the
+audio encoder, and 1, which starts the audio encoder. Valid values for
+the *channels* field are 1 (Mono audio) and 2 (Stereo audio). Valid
+values for the *codec* field are the ones sent by the server in the
+`Encoded Audio Pseudo-encoding`_ pseudo-rectangle. Valid values for the
+*kbytes-per-sec* field are codec-dependent.  The Opus codec achieves
+good performance with 32, whereas the MP3 codec might require 128 for a
+comparable experience.
+
+Encoded Audio Client Enable Audio Streaming Message
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This submessage allows the client to request the server send audio
+frames continuously. The length of an audio frame is codec-dependent,
+but is typically between 5 and 40 milliseconds.  Each frame is encoded
+with the parameters chosen by the `Encoded Audio Client Set Encoder
+State Message`_.  The client MUST send a `Encoded Audio Client Set
+Encoder State Message`_ and have received acknowledgement from the
+server that the chosen parameters are valid prior to sending this
+message.
+
+============= ======= ========= ======================================
+No. of bytes  Type    [Value]   Description
+============= ======= ========= ======================================
+1             ``U8``  245       *message-type*
+1             ``U8``  1         *submessage-type*
+2             ``U16`` 0         *payload-length*
+============= ======= ========= ======================================
+
+After invoking this operation, the client will receive a `Encoded Audio
+Server Enable Audio Streaming Message`_ with the result of the
+operation. If the operation was successful, that message will be
+followed by `Encoded Audio Server Frame Response Message`_ with and
+encoded audio frame in the corresponding container format.
+
+Once audio frames start being continuously sent, this can be stopped by
+sending a `Encoded Audio Client Set Encoder State Message`_ with the
+*enabled* field set to `0`. Due to inherent race conditions in the
+protocol, after disabling the encoder, the client may still receive
+further `Encoded Audio Server Frame Response Message`_, but once the
+server acknowledges the receipt of the `Encoded Audio Client Start
+Encoder Message`_, no further audio frames will be sent.
 
 ClientFence
 -----------
@@ -2038,6 +2137,7 @@ Number      Name
 128         Car Connectivity
 150         `EndOfContinuousUpdates`_
 173         ServerState
+245         `Encoded Audio Server Message`_
 248         `ServerFence`_
 249         OLIVE Call Control
 250         `xvp Server Message`_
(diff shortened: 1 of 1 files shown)


rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+=============== ============ ================== =======================
+
+The most significant bit of *timestamp* denotes whether the audio frame
+contains a start-of-stream header or is otherwise a keyframe, which
+enables clients to use this information for seeking purposes. Servers
+SHOULD send keyframes every few seconds / minutes to allow clients to
+re-synchronize with the stream. The 31 least significant bits of
+*timestamp* contain the number of milliseconds from the first audio
+frame that was captured in the session since the `Replit Audio Client
+Start Encoder Message`_ was acknowledged by the server. *data* SHOULD be
```
CendioOssman, line 2375: Probably easier to read if you start a new paragraph here.

rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+4               ``U32``                         *timestamp*
+*data-length*   ``U8`` array                    *data*
+=============== ============ ================== =======================
+
+The most significant bit of *timestamp* denotes whether the audio frame
+contains a start-of-stream header or is otherwise a keyframe, which
+enables clients to use this information for seeking purposes. Servers
+SHOULD send keyframes every few seconds / minutes to allow clients to
+re-synchronize with the stream. The 31 least significant bits of
+*timestamp* contain the number of milliseconds from the first audio
```
CendioOssman, line 2373: Hmm... This heavily enforces a push based model. Normally it is the output sound card that owns the clock and applications are supposed to follow.

How is this implemented in the reference server? Where does the clock come from?

rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+=============== ============ ================== =======================
+1               ``U8``       245                *message-type*
+1               ``U8``       1                  *submessage-type*
+2               ``U16``      4 + *data-length*  *payload-length*
+4               ``U32``                         *timestamp*
+*data-length*   ``U8`` array                    *data*
+=============== ============ ================== =======================
+
+The most significant bit of *timestamp* denotes whether the audio frame
+contains a start-of-stream header or is otherwise a keyframe, which
```
CendioOssman, line 2369: I'm having trouble parsing this section. If the bit is set, then it's a timestamp? And if it's not set?

rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+SHOULD send keyframes every few seconds / minutes to allow clients to
+re-synchronize with the stream. The 31 least significant bits of
+*timestamp* contain the number of milliseconds from the first audio
+frame that was captured in the session since the `Replit Audio Client
+Start Encoder Message`_ was acknowledged by the server. *data* SHOULD be
+a self-contained audio frame, and all the audio frames should be
+concatenable into a valid audio stream.  Furthermore, dropping of a
+non-keyframe SHOULD not cause the client to de-synchronize, and SHOULD
+be recoverable by inserting silence for the duration of the dropped
+frame.
```
CendioOssman, line 2380: RFB is not a lossy protocol. So is this needed?

rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+*enabled* will be set to 1 when the stream of `Replit Audio Server Frame
+Request Message`_ messages will start. *enabled* will be set to 0 if
+client had not sent a `Replit Audio Client Start Encoder Message`_
+beforehand, or if there was any other problem starting the stream. If
+there is an error at any future point, or if the client sent a `Replit
+Audio Client Start Encoder`_ with the *enabled* field set to 0, the
+server will send an additional `Replit Audio Server Start Continuous
+Updates Message`_ with *enabled* set to 0 after sending the last audio
+frame.
+
```
CendioOssman, line 2408: Video applications need to know about the latency to sync things correctly. So for them there would need to be a mechanism to report that latency as well. Have you considered such a message?

rfbproto.rst
```diff
@@ -3714,6 +3950,33 @@ The values for this pseudo-encoding are defined as follows:
 This pseudo-encoding was originally intended for use with JPEG-encoded
 subrectangles, but it could be used with other types of image encoding as well.
 
+Replit Audio Pseudo-encoding
+----------------------------
+
+A client that supports this encoding is indicating that it is able to
+receive an encoded audio data stream. If a server wishes to send encoded
+audio data, it will send a pseudo-rectangle with the following contents:
```
CendioOssman, line 3958: When is this rectangle sent? First update? Every update? Every non-incremental?

CendioOssman (changes requested):
Great, thanks. I've added some notes on things that I think could need clarification.

rfbproto.rst
lhchavez: WebRTC also has a mode of operation where the latency information is implicitly modeled at the server side and uses backpressure information in a way that doesn't require any explicit messages to be sent. This idea is used in the reference implementation since it's way simpler (and works quite well in practice).

I didn't need this message at the moment, but we can add it later if needed (hooray for versioning!), probably following the other WebRTC mode of operation where it does have an explicit latency feedback message.

rfbproto.rst
lhchavez: Done (copied it from the QEMU description, where it was also underspecified u_u).

rfbproto.rst
lhchavez: errr... it's probably useful to link to the document i'm thinking of: https://tools.ietf.org/html/draft-ietf-rmcat-gcc-02#page-4

rfbproto.rst
```diff
@@ -1404,6 +1405,124 @@ long as continuous updates are active. Non-incremental update requests
 must however be honored, even if the area in such a request does not
 overlap the area specified for continuous updates.
 
+Replit Audio Client Message
```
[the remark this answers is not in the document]
CendioOssman: Their list has lots of discrepancies compared to ours already. The important part is that the number won't get assigned to someone else. :)

rfbproto.rst
```diff
@@ -1404,6 +1405,124 @@ long as continuous updates are active. Non-incremental update requests
...
+2             ``U16``           *codec*
+2             ``U16``           *kbytes-per-sec*
+============= ======= ========= ======================================
+
+After invoking this operation, the client will receive a `Replit Audio
+Server Start Encoder Message`_ with the result of the operation.
+
+Valid values for the *enabled* field are 0, which disables/stops the
+audio encoder, and 1, which starts the audio encoder. Valid values for
+the *channels* field are 1 (Mono audio) and 2 (Stereo audio). Valid
```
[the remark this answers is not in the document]
CendioOssman: Oh? That's a bummer. I thought Opus was the go-to codec for everything. Apparently not streaming fancy video then...

rfbproto.rst
```diff
@@ -1404,6 +1405,124 @@ long as continuous updates are active. Non-incremental update requests
...
+values for the *codec* field are the ones sent by the server in the
+`Replit Audio Pseudo-encoding`_ pseudo-rect. Valid values for the
+*kbytes-per-sec* field are codec-dependent.  The Opus codec achieves
+good performance with 32, whereas the MP3 codec might require 128 for a
+comparable experience.
```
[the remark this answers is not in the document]
CendioOssman: Ah, okay.

rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+============= ======= ========= =======================================
+1             ``U8``  245       *message-type*
+1             ``U8``            *submessage-type*
+2             ``U16``           *payload-length*
+============= ======= ========= =======================================
+
+This header is then followed by arbitrary data of length
+*payload-length*, and whose format is determined by the
+*submessage-type*. Possible values for *submessage-type* and their
+associated minimum versions are
```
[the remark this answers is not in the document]
CendioOssman: So backwards compatibility largely falls to the client in that case as the server does not really know the client version. So do we need to specify versions for server-to-client messages?

We should also document the version behaviour. E.g. that the server specifies the maximum version it supports, and clients are free to use older versions.

rfbproto.rst
```diff
@@ -2171,6 +2291,121 @@ No. of bytes    Type                 [Value]    Description
...
+0.
+
+After receiveing this message with *enabled* set to 1, the client can
+send other `Replit Audio Client Message`_ messages.
+
+Replit Audio Server Frame Response Message
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This submessage contains audio data for a single audio frame wrapped in
+the container format associated with it. The length of an audio frame is
```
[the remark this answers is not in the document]
CendioOssman: I don't quite follow. Your PR to noVNC seems to merge all the frames anyway?

rfbproto.rst
CendioOssman: > Or I guess to put it in other words: would this hamper a pull-based model implementation? (I couldn't think of a way it could, but I am not an expert). Would there be any tweaks that would make a pull-based model better without breaking the push-based model? (push-based model is the one I _really_ care about since it's the only one that got good end-to-end results).

It very much would. You can really only have a single clock. So by specifying a timestamp the server is forced to provide that clock. In a pull based method the clock is at the client and is visible to the server through the frame request messages.

Right now I don't see how a pull based model can work. You would need a mechanism to disable the timestamps. E.g. they are only valid if in continuous mode. Otherwise the server should operate in a pull mode and send as much data as the client is requesting.

rfbproto.rst
CendioOssman: Ah. Do we need anything special in that case or is this really implied by the fact that the client needs to synchronise? A synchronisation will always entail either skipping some frames, or inserting fake ones, depending on which way the drift is.

rfbproto.rst
CendioOssman: I don't see any such information there right now though, so I guess watching video using this extension is currently badly out of sync?

rfbproto.rst
CendioOssman: Sorry for nit-picking, but first update after what? What if the client doesn't start with a SetEncodings message? Or if it first sends one without this extension, and then adds it?

Not extremely likely but for specifications it is usually good to be precis.

rfbproto.rst
lhchavez: wouldn't disabling the timestamp be functionally equivalent to ignoring it on the client side? there's also the consideration where some containers (like webm) require that there be timestamps in the frames (as opposed to MPEG-1, where the timestamp of the frame is implied), so even if we disable sending the timestamp, the stream will be timestamped anyways.

would adding wording along the lines of "the timestamp is informational only" help?

rfbproto.rst
lhchavez: video with audio is totally watchable! there's a very small delay depending on where the stream originates, but i've played games with this.

rfbproto.rst
lhchavez: yes, it's implied by the fact that the client needs to synchronize. reworded this a bit, since the property that needs to be upheld is that the _decoder_ stream is the one that needs to be kept valid even when dropping frames due to backpressure.

rfbproto.rst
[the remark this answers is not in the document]
lhchavez: explicit version numbers on the client messages it is then!

rfbproto.rst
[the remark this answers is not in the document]
lhchavez: there's also the backpressure implementation that is aided by this message containing individual audio frames. that way it's easier to know how many frames the client is "behind". let me add some wording about that.

rfbproto.rst
CendioOssman: The issue isn't really with the timestamp, but rather what it implies. I.e. that there is a clock on the server. So for pull to work properly that clock needs to go away and applications need to push audio data as fast as the client wants.

Any timestamps in the stream would need to be ignored in that case. Which gets me thinking, why do we have containers? They seems superfluous here, and perhaps are even a problem. Can't we send just the audio stream?

rfbproto.rst
CendioOssman: I don't quite get how that is done. Do you have some code I can look at?

rfbproto.rst
lhchavez: that's totally understandable, you *need* to be nit-picky on specifications!

rfbproto.rst
```diff
@@ -1418,7 +1418,8 @@ No. of bytes  Type    [Value]   Description
 ============= ======= ========= ======================================
 1             ``U8``  245       *message-type*
 1             ``U8``            *submessage-type*
-2             ``U16``           *payload-length*
+2             ``U16``           *client-version*
+4             ``U32``           *payload-length*
```
CendioOssman, line 1422: If the version is not part of a handshake then this is just a complicated way of extending the submessage type. So make that a U16 if you don't thing U8 is enough for future extensions. :)

rfbproto.rst
```diff
@@ -2384,13 +2388,19 @@ allow clients to re-synchronize with the stream. The 31 least
...
-of a non-keyframe due to backpressure SHOULD not cause the client to
-de-synchronize, and SHOULD be recoverable by inserting silence for the
-duration of the dropped frame.
+of a non-keyframe due to backpressure SHOULD not cause the client's audio
+stream decoder to de-synchronize. Client re-synchronization SHOULD be
+achievable by skipping frames or inserting silence frames.
+
+Server implementations MAY rely on the fact that this message contains
+audio data for a single audio frame to implement a backpressure feedback
+loop, by counting how many in-flight messages of this type exist.
```
CendioOssman, line 2403: But the server doesn't know how many of these messages the client has consumed. So how can it know what's in flight?

I think I need to see the code for this.

CendioOssman (changes requested):

rfbproto.rst
lhchavez: it can! https://github.com/replit/rfbproxy/blob/bdeae954d863bcedabb329869baad8a17e5c7a2e/src/audio/mod.rs#L184-L211

TL;DR: what this is doing is that it tries to non-blockingly send these messages as soon as it can. eventually the write will block and that's when the server knows that the client is no longer being able to accept any more messages.

rfbproto.rst
lhchavez: yup, linked it on the other thread (https://github.com/replit/rfbproxy/blob/bdeae954d863bcedabb329869baad8a17e5c7a2e/src/audio/mod.rs#L184-L211 )

rfbproto.rst
lhchavez: there's always going to be a clock on the server, but clients are always free to ignore that fact. after all, the monotonic clock is completely independent of the audio subsystem.

> why do we have containers?

i'm also trying to simplify the client implementation: having them have to containerize the audio stream pushes a lot of complexity to their side, where they don't necessarily have the libraries for it (e.g. noVNC, the use case I really care about).

there's also the fact that opus purposefully requires there to be a container (since the stream is not self-delimited), and they recommend the container to be vorbis or webm instead of rolling a hand-crafted one.

rfbproto.rst
lhchavez: oh you know whatr? the Set Encoder State *is* the handshake, so we can only add it there.

rfbproto.rst
CendioOssman: That just tells you some buffer is full. You don't know the size of that buffer. Or any buffering beyond that. And I don't see the latency actually reported back to the applications?

I would suspect you just have such low latency in the system in your test case that the issue isn't apparent.

You mentioned you saw a large latency in the browser? How did you see that?

rfbproto.rst
CendioOssman: > there's always going to be a clock on the server, but clients are always free to ignore that fact. after all, the monotonic clock is completely independent of the audio subsystem.

In a pull model such a thing is not allowed. Applications can just follow a single clock, and that clock would need to be the client's in a pull model.

Having had a look at your code I'd say the clock is in PulseAudio or the sound card on the server. I assume you attach to the monitor of the default sink on the server. In this method PulseAudio remains in control of the clock. TBH, I'm not sure if there is a way for a normal PulseAudio client to own the clock. You probably need to be a PulseAudio module for that.

So given your implementation a pull model is impossible. So it might be best to make this extension push only and leave pull for the next guy.

> i'm also trying to simplify the client implementation: having them have to containerize the audio stream pushes a lot of complexity to their side, where they don't necessarily have the libraries for it (e.g. noVNC, the use case I really care about).

I was hoping the clients could make use of the raw stream directly. If that isn't the case then I understand the need for a container.

lhchavez: > So given your implementation a pull model is impossible. So it might be best to make this extension push only and leave pull for the next guy.

works for me! in the end the push model was the one i *really* care about.

rfbproto.rst
lhchavez: > That just tells you some buffer is full. You don't know the size of that buffer. Or any buffering beyond that. And I don't see the latency actually reported back to the applications?

Yeah, but all of that is sufficient to get a decent experience.

> I would suspect you just have such low latency in the system in your test case that the issue isn't apparent.

I wouldn't discount this being the case, but I am still able to experience _decent_ (not amazing) performance when running in a constrained system (a container with CPU scheduling limits in a shared VM) over the internet.
 
> You mentioned you saw a large latency in the browser? How did you see that?

In the noVNC patch, if i tweak this constant https://github.com/novnc/noVNC/pull/1525/files#diff-16e913176fa58949bba328b2391e36efaddee77167582a2b4e4eeb33c96e5be4R20 and try to lower it more, the browser just pauses the audio in order to re-buffer. and afterwards, it _tries_ to keep ~300msec worth of buffer.

I added some console.logging with the position of the stream, the timestamps of the incoming frames, and the internal positions of the AudioBuffer to figure all that out.

CendioOssman: Do you have instructions for your proxy on how to set things up for testing?

lhchavez: welp! completely forgot to do that. I'll add it to the README soon, but for the time being,

```shell
Xtigervnc -rfbport=5901 -SecurityTypes=None --localhost :0 &
cd rfbproxy && cargo run -- --enable-audio
```

rfbproxy needs to have a Rust toolchain (https://rustup.rs/) installed, plus the libopus-dev libpulse-dev libmp3lame-dev packages.

rfbproto.rst
```diff
@@ -1047,6 +1047,7 @@ Number      Name
 127         VMware
 128         Car Connectivity
 150         `EnableContinuousUpdates`_
+245         `Replit Audio Client Message`_
```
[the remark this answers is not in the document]
samhed: Any news here?

samhed:
What's the status on this one, has all the comments been addressed?

rfbproto.rst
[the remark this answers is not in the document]
lhchavez: no :( let me try pinging them.

CendioOssman:
Setting up a test environment is still on my todo so I can get a better picture of the latency feedback situation.

lhchavez:
> Setting up a test environment is still on my todo so I can get a better picture of the latency feedback situation.

you can use https://pulseaudio.luisreplit.repl.co/ as test environment.

rfbproto.rst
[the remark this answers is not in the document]
lhchavez: the assignments are live! https://www.iana.org/assignments/rfb/rfb.xhtml#Luis_H%C3%A9ctor_Ch%C3%A1vez

rfbproto.rst
CendioOssman: So I got this up and running now. And I can easily tweak that value and create massive buffering on the client that the server has no knowledge about. So this section should be removed, or at the very least be changed to indicate that the only feedback that exists is any provided by the transport layer.

The biggest issue without a good feedback system is normally that you overfill some buffers. Fortunately this use case should have a steady data rate so as long as bandwidth is sufficient that should not happen.

Instead it is crucial that the client detects synchronisation issues and either drops data or inserts it to sync back up. I think it would be helpful if this is very clear in the specification. I've seen systems in the past where this was overlooked.

The client is also responsible for maintaining sufficient buffering to handle network jitter. And since there is no latency feedback mechanism the client should be advised to try to maintain a minimal buffer.

Latency in practice is definitely below 300 ms. However I'm getting choppy sounds so something is getting underruns. It doesn't seem to be the client so I'm guessing it's in the proxy somewhere.

lhchavez: that all sounds reasonable! added a lot of clarification for that.

rfbproto.rst
```diff
@@ -1404,6 +1405,105 @@ long as continuous updates are active. Non-incremental update requests
...
+good performance with 32, whereas the MP3 codec might require 128 for a
+comparable experience.
+
+Encoded Audio Client Start Continuous Updates Message
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This submessage allows the client to request the server send audio
+frames continuously, which saves bandwidth and reduces audio latency
+incurred by the TCP stack by half compared to requesting frames
+individually. The length of an audio frame is codec-dependent, but is
```
CendioOssman, line 1478: Frames can no longer be requests individually so this wording needs to be tweaked.

Perhaps also rename the message?

rfbproto.rst
```diff
@@ -2171,6 +2272,142 @@ No. of bytes    Type                 [Value]    Description
...
+recommended to maintain a minimal buffer of at least three audio frames.
+
+Server implementations MAY rely on the fact that this message contains
+audio data for a single audio frame to implement a backpressure feedback
+loop, by counting how many messages of this type are waiting to be
+delivered through the transport layer to the client. Using this
+information, the server can decide to drop audio frames if a client lags
+behind too much, by consuming the audio frames from the audio source and
+not advancing the encoder state. This way, the stream can still be
+decoded by the client.
```
CendioOssman, line 2382: Do you see a need for keeping this? I think it's enough that we've stated that the client should handle the synchronisation.

CendioOssman (changes requested):
Some cleanup from all the changes and we should be good to go.

rfbproto.rst
lhchavez: yeah, it's a good suggestion to give server implementers: since the client is responsible for synchronization, it's something that servers can do without issues.

the reference implementation also uses it, so i would prefer for that behavior to be present in the spec.

rfbproto.rst
lhchavez: Removed the reference to request frames individually. 

Any suggestions for the message name? "Start Continuous Updates" is kinda consistent with the "ContinuousUpdates" message, which is why that was chosen. If it weren't for the consistency argument, I would have chosen "Enable Audio Streaming", so I tentatively went with that.

lhchavez:
rebased and squashed the commits

lhchavez:
gentle ping. is there any chance this could be merged?

CendioOssman:
I'm afraid I haven't had time to continue my testing. Since there was an issue there I'd like to hold off on merging this in case there is a protocol problem. It's still on my todo though and I hope to get back to it and get all of this resolved.

lhchavez:
> I'm afraid I haven't had time to continue my testing. Since there was an issue there I'd like to hold off on merging this in case there is a protocol problem. It's still on my todo though and I hope to get back to it and get all of this resolved.

is there anything i can help with? the protocol as is written has been running in production for a while with no issues.

CendioOssman:
I think you wrote some suggestions on the noVNC PR. I just need to find time to try them out.

mdevaev:
Hello! I'm also interested in this, but I have a couple of questions..
* I watched the discussion, but I still didn't understand why it was necessary to put OPUS frames in a WebM container. There is a similar RTP specification that allows you to transmit OPUS audio, and RAW frames are used there. It seems to me that the need for WebM is due to the browser requirements for noVNC, because without WebM it will not be able to play OPUS, am I right?
* You have provided a message that allows the server to tell which codecs it supports. But I would like to clarify the following behavior: what happens if the client has chosen a supported codec, but the encoding parameters do not match those supported by the server, or the server is not able to change them at all? I use a single OPUS encoder process and don't want to let users change the settings. That is, the server is like, "You asked for OPUS, okay, here's OPUS for you, but the parameters will be different, it will be stereo and 48k, not mono and 44.1k". I suggest either expanding the server response so that it contains the settings, and the client can understand whether the server has accepted the requested settings or can give the audio, but with others. Or fix the behavior of the server in this specification, that the bitrate and channel settings for the selected codec are a preference, and the server can ignore them.

Neustradamus:
Any progress on it?

PR #44, opened 2021-11-25
Add JPEG encoding

pdlan:
This documentation of the JPEG encoding comes from my blackbox analysis of the network traffic between RealVNC servers and clients. I'm not completely sure about its correctness but my own server-side implementation works quite well with RealVNC clients. Also, I'm sure the PR (https://github.com/novnc/noVNC/pull/1609) of noVNC works with (at least latest) RealVNC servers.

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -2362,6 +2362,7 @@ Number       Name
 7            `Tight Encoding`_
 8            `zlibhex Encoding`_
 16           `ZRLE Encoding`_
+21           `JPEG Encoding`_
 50           `Open H.264 Encoding`_
 -260         `Tight PNG Encoding`_
 ============ ==========================================================
@@ -2410,7 +2411,6 @@ Number                      Name
 15                          TRLE
 17                          Hitachi ZYWRLE
 20                          H.264
-21                          JPEG
 22                          JRLE
 1000 to 1002                Apple Inc.
 1011                        Apple Inc.
@@ -3113,6 +3113,25 @@ are:
     =================== =============== ======= =======================
 
     Where *r* is floor((*runLength* - 1) / 255).
+
+JPEG Encoding
+-------------
+
+The JPEG encoding uses JPEG format to compress rects. The message is
+simply a JPEG image:
+
+=============== ================ ======================================
+No. of bytes    Type             Description
+=============== ================ ======================================
+variable        ``U8`` array     *data*
+=============== ================ ======================================
+
+As defined in the JPEG standard, a JPEG image consists of a sequence of
+segments, each of which begins with a marker. The EOF marker indicates
+the end of the message. When the segments of
+"Define Huffman Tables (DHT)" or "Define Quantization Tables (DQT)"
+do not exist, the client should reuse the previous Huffman tables or 
+quantization tables for decoding.
 
 Open H.264 Encoding
 -------------------


rfbproto.rst
```diff
@@ -3114,6 +3114,25 @@ are:
...
+-------------
+
+The JPEG encoding uses JPEG format to compress rects. The message is
+simply a JPEG image:
+
+    =============== ================ ==================================
+    No. of bytes    Type             Description
+    =============== ================ ==================================
+    variable        ``U8`` array     *data*
+    =============== ================ ==================================
```
CendioOssman, line 3127: General tables shouldn't be indented. It's only when ther are part of a list or such.

rfbproto.rst
```diff
@@ -3114,6 +3114,25 @@ are:
...
+
+As defined in the JPEG standard, a JPEG image consists of a sequence of
```
CendioOssman, line 3129: Nitpicking a bit, but JPEG is just the algorithm. I'm assuming the format is JFIF, which might be useful to specify.

I don't suppose you know if arithmetic coding is allowed? It was commonly left out since it used to have a patent obstacle.

CendioOssman (changes requested):
Many thanks.

We like to have signed off-lines for our commits though. Could you adjust your commit message and add that? git can add it for you using the `-s` line.

rfbproto.rst
pdlan: I think there is a JPEG standard (ITU T.81) that defines the format of a JPEG image (segments, markers, etc.) while JFIF defines what segments and metadata should appear in that file. So I think it's ok to simply say a JPEG image as I'm not sure if it's limited to JFIF (or Exif?, an alternative JPEG standard which is incompatible with JFIF) or not. As for you second question, I have no idea whether arithmetic coding is allowed. But from my observasion it seems that RealVNC servers don't use it.

CendioOssman: I'm unfortunately not familiar with the details, so I'm going on what Wikipedia and such sources are claiming. And I'm interpreting the information there to say that the JPEG standard does not have a complete definition of the format, so things like JFIF or EFIF are needed:

http://fileformats.archiveteam.org/wiki/JPEG
http://fileformats.archiveteam.org/wiki/JFIF

The above page also seems to say that libjpeg is just JFIF (not EXIF), and libjpeg is pretty much the de facto standard.

I'm mentioning this out of a slight concern of interoperability issues if someone decides to add more fancy stuff in to the JPEG stream. The specification should be independent and not defined by whatever RealVNC's implementation happens to accept.

CendioOssman: What have you seen in practice? The JFIF and EXIF formats seem to have headers clearly marking which format is used.

pdlan: It seems that the RealVNC server uses neither. I just captured some packets and found that the JPEG data doesn't even have a APP0/1 segment, which is required for JFIF or EXIF.

CendioOssman: Ouch. But I guess libjpeg gladly accepts the data anyway?

pdlan: Yes, it's fine if only some metadata is missed. However, if the segments of quatization or Hufman tables are ignored, no image decoder can handle that situation, so we have to cache the tables and insert them into the image for a normal decoder to decode it (as I did in my noVNC's PR).

CendioOssman: Alright. I'll guess we can leave it a bit vague then until someone does more experimentation around the finer details.

CendioOssman (commented):
Please also extend the table so it has the same width as the rest. Then once you've squashed the commits I can go ahead and merge this. Thanks.

CendioOssman (changes requested):
The commits look correctly squashed, but please review the commit message so it only contains the essentials. Also don't forget about the signed off line in the commit message.

CendioOssman:
I'm afraid the signed off line is still missing. Please run `git commit --amend -s` and it should be added for you.

PR #45, opened 2021-11-28
Add RA2, RA2ne, RA2r, RA2_256, RA2ne_256, RA2r_256 security type

pdlan:
It's an initial documentation of RA2 security types for my previous PR https://github.com/novnc/noVNC/pull/1351. Sorry for the late of nearly 2 years.

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -372,10 +372,16 @@ Number      Name
 0           Invalid
 1           `None`_
 2           `VNC Authentication`_
+5           `RSA-AES Security Type`_
+6           `RSA-AES Unencrypted Security Type`_
+13          `RSA-AES Two-step Security Type`_
 16          `Tight Security Type`_
 19          `VeNCrypt`_
 22          `xvp Authentication`_
 30          `Diffie-Hellman Authentication`_
+129         `RSA-AES-256 Security Type`_
+130         `RSA-AES-256 Unencrypted Security Type`_
+133         `RSA-AES-256 Two-step Security Type`_
 =========== ===========================================================
 
 Other registered security types are:
@@ -384,9 +390,8 @@ Other registered security types are:
 Number      Name
 =========== ===========================================================
 3-4         RealVNC
-5           RA2
-6           RA2ne
-7-15        RealVNC
+7-12        RealVNC
+14-15       RealVNC
 17          Ultra
 18          TLS
 20          SASL
@@ -394,7 +399,9 @@ Number      Name
 23          Secure Tunnel
 24          Integrated SSH
 31-35       Apple Inc.
-128-255     RealVNC
+128         RealVNC
+131-132     RealVNC
+134-255     RealVNC
 =========== ===========================================================
 
 The official, up-to-date list is maintained by IANA [#reg]_.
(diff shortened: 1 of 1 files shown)


rfbproto.rst
```diff
@@ -372,8 +372,14 @@ Number      Name
...
 1           `None`_
 2           `VNC Authentication`_
+5           `RA2 <RA2 Security Type_>`_
+6           `RA2ne <RA2 Security Type_>`_
+13          `RA2r <RA2 Security Type_>`_
 16          `Tight Security Type`_
 19          `VeNCrypt`_
+129         `RA2_256 <RA2 Security Type_>`_
+130         `RA2ne_256 <RA2 Security Type_>`_
+133         `RA2r_256 <RA2 Security Type_>`_
```
CendioOssman, line 382: It would be preferred if this could match the actual heading it refers to. Have you tried some other description? E.g. "RA2 Security Type (RA2r)"?

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
 ..
   XXX: Correct link to the SASL method when it gets accepted.
 
+RA2 Security Type
+-----------------
+RA2, RA2ne, RA2r, RA2_256, RA2ne_256, RA2r_256 security types are similar.
+Only symmetric key sizes and hash algorithms are different. So we document
+them together.
```
CendioOssman, line 816: This is probably very subjective, but I can find this gets a bit messy when all are handled at the same time and you get a exceptions in the text in multiple places.

Did you see how things would look if you handled them more separately? E.g.

```
RA2ne Security type
-------------------

The RA2ne security type is identical to the RA2 security type, except that only the
security handshake is encrypted. The SecurityResult message and all following data
remains unencrypted.
```

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+
+After these security types are selected, the server sends its RSA public key
+of 2048 bits. The public key, i.e., the modulus and the public exponent, are
+big-endian integers.
+
+=============== ============ ======= ==================================
+No. of bytes    Type         Value   Description
+=============== ============ ======= ==================================
+4               ``U32``      2048    key length in bits
```
CendioOssman, line 825: Since the size is explicitly specified I assume this can vary? E.g. this man page says that 2048 is only the default:

https://archive.realvnc.com/products/vnc/documentation/4.3/unix/man/vnckeygen.html

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+=============== ============ ======= ==================================
+
+The server's random number is decrypted with the client's private key. At the
+same time, the client's random number is decrypted with the server's private
+key. Now both random numbers are known by both sides. The client session key
+and server session key can be derived as follows:
+
+For RA2, RA2ne, RA2r:
+
+::
```
CendioOssman, line 868: FYI, you can include those `::` at the end of the preceding paragraph and it will get the same formatting.

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+
+    ClientSessionKey = SHA256(ServerRandom || ClientRandom)
+    ServerSessionKey = SHA256(ClientRandom || ServerRandom)
+
+The client session key is used to encrypted the messages from the client to
+the server. The server session key is used to encrypted the messages from
+the server to the client.
+
+After that, all the messages will be encrypted with the AES-EAX mode, which
+involves AES-CTR and CMAC. For each message, there is a message header of a
```
CendioOssman, line 887: Nice that you managed to figure out what the algorithm was!

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+The plaintext message of the credentials is:
+
+===================== ============ ====================================
+No. of bytes          Type         Description
+===================== ============ ====================================
+1                     ``U8``       *length-username* (is 0 if subtype=2)
+*length-username*     ``U8`` array username (can be empty)
+1                     ``U8``       *length-password*
+*length-password*     ``U8`` array password
+===================== ============ ====================================
```
CendioOssman, line 980: Have you figured out the encoding of these? UTF-8? ISO 8859-1?

CendioOssman (changes requested):
Nice! Thanks!

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+
+=============== ============ ======= ==================================
+No. of bytes    Type         Value   Description
+=============== ============ ======= ==================================
+4               ``U32``      2048    key length in bits
+256             ``U8`` array         modulus
+256             ``U8`` array         public exponent
+=============== ============ ======= ==================================
+
+Then the client also sends its public key.
```
CendioOssman, line 830: I think it is important to mention that the client has to verify the server's key. Otherwise there is little real security provided here.

CendioOssman (changes requested):

rfbproto.rst
```diff
@@ -805,6 +809,189 @@ Authentication continues with the SASL method when TLS handshake is completed.
...
+
+=============== ============ ======= ==================================
+No. of bytes    Type         Value   Description
+=============== ============ ======= ==================================
+4               ``U32``      2048    key length in bits
```
CendioOssman, line 835: Can the client have a key that is of a different size than the server? Give that it is a long lived key I can hope that is the case.

And what happens if the key is too short? Or too long? Does the server send some special message?

CendioOssman (changes requested):

rfbproto.rst
```diff
@@ -372,8 +372,14 @@ Number      Name
 0           Invalid
 1           `None`_
 2           `VNC Authentication`_
+5           `RA2 <RA2 Security Type_>`_
```
CendioOssman, line 375: More descriptive names are also preferred, now that we know how these work. I would guess the "RA2" is short for "RSA Authentication".

CendioOssman (changes requested):

rfbproto.rst
pdlan: From [this document](https://static.realvnc.com/media/documents/realvnc-rfb-protocol-security-analysis.pdf) it seems that RA2 stands for "RSA with AES version 2". I will use the name "RSA-AES Authentication" instead.

rfbproto.rst
pdlan: Yes I confirmed that the sizes can be different.

rfbproto.rst
pdlan: Yes, I tried to use the keys of the sizes from 512 bits to 8192 bits and the RealVNC client accepted all of them.

rfbproto.rst
CendioOssman: Sounds good! :+1:

That document also mentions an RA4. Anything you've looked at?

rfbproto.rst
```diff
@@ -372,10 +372,16 @@ Number      Name
 0           Invalid
 1           `None`_
 2           `VNC Authentication`_
+5           `RSA-AES Security Type (RA2) <RSA-AES Security Type_>`_
+6           `RSA-AES Unencrypted Security Type (RA2ne) <RSA-AES Unencrypted Security Type_>`_
+13          `RSA-AES Two-step Security Type (RA2r) <RSA-AES Two-step Security Type_>`_
```
CendioOssman, line 377: We don't need to keep the short forms now that they have different heading. Avoids having to have the `<>` syntax.

rfbproto.rst
```diff
@@ -384,8 +390,6 @@ Other registered security types are:
 Number      Name
 =========== ===========================================================
 3-4         RealVNC
-5           RA2
-6           RA2ne
```
CendioOssman, line 388: 129, 130 and 133 should also be removed from the list.

rfbproto.rst
```diff
@@ -810,6 +814,207 @@ completed.
 ..
   XXX: Correct link to the SASL method when it gets accepted.
 
+RSA-AES Security Type
+---------------------
+After these security types are selected, the server sends its RSA
+public key. The public key, i.e., the modulus and the public exponent,
+are big-endian integers. If the key length is too small or too big,
```
CendioOssman, line 821: I guess they are unsigned as well since there is no padding for an extra leading 0?

CendioOssman (changes requested):

rfbproto.rst
CendioOssman: Ah, sorry. I missed that some things were still grouped.

I think splitting is fine. This is a reference documentation, so some things naturally get a bit repetitive. :)

I assume these extra chapters are extremely short? It was just a swap of SHA-1 for SHA-256, right?

rfbproto.rst
pdlan: From my observation RealVNC can only use RA2 with RFB3.8. RA4 is used for RFB4+, which is not documented at all.

CendioOssman (changes requested):
Looks great! Thanks!

Please squash the commits, finalize the commit message and add a signed-off line and I'll go ahead and merge this.

pdlan:
Hi @CendioOssman, any update on this?

CendioOssman:
Sorry, I missed that your last update (not sure if GitHub sent a notification?). I'll have a look now.

pdlan:
Great!I think GitHub might also have not sent the notification of the noVNC's PR. You could also have a look at that if it's convenient for you.

PR #54, opened 2023-10-01
Document PTS and NTP

any1:
These extensions enable smooth rendering based on presentation time stamps and NTP-like time synchronisation.

I've been working on presentation time based frame scheduling for Neat VNC and wlvncc. This documents extensions that are necessary to implement such frame scheduling.

The implementation is complete w.r.t. the server side, but the client side is at a proof-of-concept stage. See https://github.com/any1/neatvnc and https://github.com/any1/wlvncc/tree/pts-frame-sched

Points of interest in the code:
- Frame selection: https://github.com/any1/wlvncc/blob/pts-frame-sched/src/main.c#L785
- NTP: https://github.com/any1/wlvncc/blob/pts-frame-sched/src/ntp.c

These extensions also happen to be useful to profiling and tracking frames between the server and client and measuring end-to-end latency. See: https://andri.yngvason.is/measuring-latencies.html

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -2722,6 +2722,8 @@ Number       Name
 -314         `Cursor With Alpha Pseudo-encoding`_
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
 -763 to -768 `JPEG Subsampling Level Pseudo-encoding`_
+-1000        `Presentation time Pseudo-encoding`
+-1001        `NTP Pseudo-encoding`
 0x574d5664   `VMware Cursor Pseudo-encoding`_
 0x574d5665   `VMware Cursor State Pseudo-encoding`_
 0x574d5666   `VMware Cursor Position Pseudo-encoding`_
@@ -4489,3 +4491,72 @@ be set.
     4               ``U32``                         *size*
     *size*          ``U8`` array                    *data*
     =============== ==================== ========== ===================
+
+Presentation Time Pseudo-Encoding
+---------------------------------------
+
+A client that requests the Presentation Time pseudo-encoding is
+declaring that it wants to receive presentation timestamps with each
+frame.
+
+If the server supports presentation time, each frame from the server
+must contain a PTS pseudo-rectangle which contains the time at which
+the frame was presented to the user in micro seconds.
+
+The contents of the pseudo-rectangle are:
+
+=============== =================== ===================================
+No. of bytes    Type                Description
+=============== =================== ===================================
+4               ``U32``             *high bits*
+4               ``U32``             *low bits*
+=============== =================== ===================================
+
+NTP Pseudo-Encoding
+---------------------------------------
+
+A client that requests the NTP pseudo-encoding is declaring that it
+supports time synchonisation with the server akin to NTP. To indicate
+that the server also supports NTP messages, the server must send an
+empty rectangle with the encoding type set to -1001.
+
+This extension is useful in combination with the Presentation Time
+pseudo-encoding. It can be used to estimate jitter and to translate
+timestamps originating from the server to the time domain of the
+client. The client can then attempt to render frames using the original
+time interval, smoothing out frame presentation on jittery networks.
+
+NTP events have the event type 160 and they have the same size and
+structure, regardless of whether they come from the server or the
+client.
+
+All timestamps are in micro seconds.
+
+=============== =================== ===================================
+No. of bytes    Type                Description
+=============== =================== ===================================
+1               ``U8``              *type*
+3               ``U8``              *padding*
+4               ``U32``             *t0*
+4               ``U32``             *t1*
+4               ``U32``             *t2*
+4               ``U32``             *t3*
+=============== =================== ===================================
+
+To find out the server's clock relative to the client's, the client
+must periodically ping the server with an NTP message. The client must
+set `t0` to its own current time and all other timestamps should be set
+to zero.
+
+When the server receives the message, it must set `t1` to the time at
+which the message was received using its own current time. The server
+should the same message that it received to be transmitted back to the
+client. Rigth before the message is transmitted to the network, `t2`
+should be set to the current time.
+
+Manual output queuing with a minimal OS-side send queue size is
+recommended to minimise error due ot HOL blocking.
+
+After the client receives the message with `t1` and `t2`, it should set
+`t3` and send the message back to the server. This way, both parties
+can know the time difference.


CendioOssman:
Thanks for your contribution.

I'm not entirely clear on the purpose, though. The description here talks about rendering, but the blog post seems to be entirely focused on measurements?

any1:
Like I said: These extension *also* happen to be useful for measurements. I haven't written a blog post yet about frame scheduling.

The main utility is smooth rendering.

In streaming protocols such as RTP, a timestamp is usually attached to every frame. This is the time at which the frame is supposed to be rendered in order for playback to be smooth. This implements the same for RFB.

This is *also* useful for synchronising video and audio, so if (for some reason), you want to watch a video over VNC, you would be able to watch that video without the audio or video lagging behind. I haven't implemented it, but timestamps are needed for it to work.

WebRTC uses timestamps in combination with a jitter buffer to smooth out frame presentation. This offers us the ability to do the same. Using the NTP extension, jitter can be measured. Based on the jitter, we can calculate how much we need to add to the presentation deadline for each frame. This increases the average presentation latency, but not the maximum presentation latency.

rfbproto.rst
```diff
@@ -2722,6 +2722,8 @@ Number       Name
 -314         `Cursor With Alpha Pseudo-encoding`_
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
 -763 to -768 `JPEG Subsampling Level Pseudo-encoding`_
+-1000        `Presentation time Pseudo-encoding`
+-1001        `NTP Pseudo-encoding`
```
CendioOssman, line 2726: Is a registration with IANA for these pending? I couldn't see anything in the current allocations.

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
     4               ``U32``                         *size*
     *size*          ``U8`` array                    *data*
     =============== ==================== ========== ===================
+
+Presentation Time Pseudo-Encoding
+---------------------------------------
```
CendioOssman, line 4496: We've tried to have the underline match the heading for these.

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
...
+the frame was presented to the user in micro seconds.
+
+The contents of the pseudo-rectangle are:
+
+=============== =================== ===================================
+No. of bytes    Type                Description
+=============== =================== ===================================
+4               ``U32``             *high bits*
+4               ``U32``             *low bits*
+=============== =================== ===================================
```
CendioOssman, line 4513: I don't suppose we should just define a `U64`?

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
...
+NTP Pseudo-Encoding
+---------------------------------------
+
+A client that requests the NTP pseudo-encoding is declaring that it
+supports time synchonisation with the server akin to NTP. To indicate
+that the server also supports NTP messages, the server must send an
+empty rectangle with the encoding type set to -1001.
+
+This extension is useful in combination with the Presentation Time
+pseudo-encoding. It can be used to estimate jitter and to translate
```
CendioOssman, line 4524: A link here perhaps?

CendioOssman, line 4524: I guess the idea is that these two extensions must use the same clock? Perhaps that should be explicit?

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
...
+Presentation Time Pseudo-Encoding
+---------------------------------------
+
+A client that requests the Presentation Time pseudo-encoding is
+declaring that it wants to receive presentation timestamps with each
+frame.
+
+If the server supports presentation time, each frame from the server
+must contain a PTS pseudo-rectangle which contains the time at which
+the frame was presented to the user in micro seconds.
```
CendioOssman, line 4504: What kind of clock must it be? Guaranteed monotonic? Or can you use any clock?

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
...
+
+This extension is useful in combination with the Presentation Time
+pseudo-encoding. It can be used to estimate jitter and to translate
+timestamps originating from the server to the time domain of the
+client. The client can then attempt to render frames using the original
+time interval, smoothing out frame presentation on jittery networks.
+
+NTP events have the event type 160 and they have the same size and
+structure, regardless of whether they come from the server or the
+client.
```
CendioOssman, line 4531: Please document this in the messages section. Ideally under both client and server sections if they can be sent in both direction. You can look at the fence messages for an existing example.

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
...
+should the same message that it received to be transmitted back to the
+client. Rigth before the message is transmitted to the network, `t2`
+should be set to the current time.
+
+Manual output queuing with a minimal OS-side send queue size is
+recommended to minimise error due ot HOL blocking.
+
+After the client receives the message with `t1` and `t2`, it should set
+`t3` and send the message back to the server. This way, both parties
+can know the time difference.
```
CendioOssman, line 4562: I assume these different timestamps are things from the NTP specification? It is not obvious how all of these should be handled.

If it uses the same system as NTP, then perhaps we can reference appropriate sections in that to avoid duplication?

rfbproto.rst
```diff
@@ -4489,3 +4491,72 @@ be set.
...
+To find out the server's clock relative to the client's, the client
+must periodically ping the server with an NTP message. The client must
+set `t0` to its own current time and all other timestamps should be set
+to zero.
+
+When the server receives the message, it must set `t1` to the time at
+which the message was received using its own current time. The server
+should the same message that it received to be transmitted back to the
+client. Rigth before the message is transmitted to the network, `t2`
+should be set to the current time.
```
CendioOssman, line 4555: Won't `t1` and `t2` be the same if the server tries to respond immediately?

rfbproto.rst
CendioOssman: They use different types, though. So I guess they cannot be the same clock?

CendioOssman (changes requested):

any1:
> Given that VNC is an interactive system, I would have guessed that users prefer responsiveness to smoothness.
>

That depends on what they're doing. In any case, I would argue that the latency is only as good as the maximum latency. With this, the latency becomes more consistent while not exceeding what would otherwise be the maximum latency by much.

> Have you thought about guidelines for how clients should best implement this to get a good user experience?

We'll see when I'm further along with implementing this myself. It's at a proof-of-concept stage currently. I just wanted to get this out there in order to get some feedback. Perhaps I ought to turn this PR into a Draft.

Current guideline is, roughly:
- Ping the server regularly using NTP
- Translate the server's PTS to the client's clock using NTP
- Get the max jitter from NTP
- Add it to the presentation deadline for the frames
- Render frames as close to the presentation deadline as possible

rfbproto.rst
any1: No, should there be? I haven't submitted anything to IANA before.

rfbproto.rst
any1: They are the same clock, but when they're used together, the upper bits of pts are dropped. Perhaps I ought to just make them all u64 or all u32.

u32 is enough for what I want to do with it, but someone else might want to do something that requires u64. So, I'm conflicted...

rfbproto.rst
any1: Monotonic.

rfbproto.rst
any1: Correct. However, they needn't be the same. In Neat VNC, I implemented a send queue in user space and set SNDBUF to a smaller value than the default. When an NTP message is taken from the internal send queue, `t2` is set before the message is finally sent to the OS's send queue.

CendioOssman:
@any1, this something you are still working on?

any1:
I am not currently working on this, but I haven't abandoned this either.

Neustradamus:
Any progress on it?

PR #55, opened 2023-12-17
Document Unicode keysyms

qmfrederik:
Keysym values in the 0xFF00 - 0xFFFF range conflict with Unicode characters in the same range.

This commit adds a note indicating the presence of [Unicode keysyms](https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#Unicode_KEYSYM). They are:
- Part of [recent versions of the X Window System Protocol](https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#Unicode_KEYSYM)
- Documented in [RFC 6143](https://datatracker.ietf.org/doc/html/rfc6143#section-7.5.4)
- Appear to be supported by at least [TurboVNC](https://github.com/TurboVNC/turbovnc/blob/33550c8453235c99c9fadc659eb56739c7857bb9/unix/vncviewer/keysym2ucs.c#L866) and [novnc](https://github.com/novnc/noVNC/blob/829725b30e3d3486991e34db0c86406e556ebf98/core/input/keysymdef.js#L686)

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1643,6 +1643,12 @@ used:
   viewers, servers should also recognise ISO Left Tab as meaning a
   shifted Tab.
 
+- Unicode characters can be represented using Unicode keysyms, which
+  occupy the range `0x01000100`` to `0x0110FFFF`` and represent the ISO
+  10646 / Unicode characters `U+0100` to `U+10FFFF`, respectively. The
+  numeric value of a Unicode keysym is the Unicode position of the
+  corresponding character plus `0x01000000`.
+
 Note that extensions such as `QEMU Extended Key Event Message`_ provide
 alternative behaviours for keyboard events that do not follow what is
 described here.


CendioOssman:
Thank you for your contribution!

That section is not meant to be a full documentation on keysyms, but rather behaviour that is somewhat unique to RFB/VNC.

The addition suggested here seems to just document the standard behaviour for modern X servers?

qmfrederik:
Thanks @CendioOssman .  Yes, this does seem to be the standard behavior for modern X servers.  It took me quite some time to find out how Unicode characters could be represented, so that's why I thought it'd be worth at least dedicating a paragraph to support for Unicode.

Alternatively, another option is to amend the fourth paragraph in the KeyEvent section section, leaving a note about Unicode (since it already mentions ASCII) and referencing the X Window System Protocol instead of the O'Reilly book?

A proposed change would be:

_For most ordinary keys, the "keysym" is the same as the corresponding ASCII value. **Unicode characters are represented by their Unicode position plus 0x01000000**.  For full details, **see [X Window System Protocol](https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#keysym_encoding)**, or see the header file <X11/keysymdef.h> from any X Window System installation. Some other common keys are:_

CendioOssman:
The O'Reilly book and the link you provided are the same thing. The original text is just too old for anyone to have a good online link for it. So that seems like an obvious improvement in usability, yes. :)

The suggested text is easily misunderstood as recommending using the Unicode range for everything outside of ASCII. I think this is a very bad idea, as those keysyms are not what most systems are using. The norm is to use what the X specification refers to as "Legacy keysyms".

I see that RFC 6143 has a wording just along those lines. So if we want to add something, then something very similar to that would probably be best.

PR #57, opened 2024-07-03
Document ExtendedMouseButtons pseudo-encoding

CendioHalim:
This pseudo-encoding adds support for additional mouse buttons, most notably the forward and backward mouse buttons.
* Open PR: https://github.com/TigerVNC/tigervnc/pull/1826

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1967,6 +1967,76 @@ The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
 negotiation of an alternative interpretation for the *x-position*
 and *y-position* fields, as relative deltas.
 
+The `ExtendedMouseButtons Pseudo-encoding`_ extends the mouse button
+functionality by adding an additional byte for button states. When
+this extension is enabled, the highest bit in the *button-mask*
+can be set to one, which marks the `PointerEvent`_ message as an
+extended `PointerEvent`_ message and replaces bit used for the
+back button.
+
+=============== ==================== ========== =======================
+No. of bytes    Type                 [Value]    Description
+=============== ==================== ========== =======================
+1               ``U8``               5          *message-type*
+1               ``U8``                          *button-mask*
+2               ``U16``                         *x-position*
+2               ``U16``                         *y-position*
+1               ``U8``                          *extended-buttons-mask*
+=============== ==================== ========== =======================
+
+The extended `PointerEvent`_ message provides support for 7 additional
+buttons, giving a total of 15 buttons. To allow additional data being
+transmitted using the same `PointerEvent`_ message, a marker bit is
+used to indicate that there is an extra byte of data in the
+`PointerEvent`_ message. The marker bit is the highest bit of the
+*button-mask*.
+
+An extended `PointerEvent`_ adds an *extended-buttons-mask* to the
+`PointerEvent`_ message, containing the button state for additional
+mouse buttons. The *button-mask* will contain the button state for
+buttons 1-7 defined in a normal `PointerEvent`_ message.
+
+Clients that support extended `PointerEvent`_ messages must always set
+the marker bit to zero until the server indicates that it also supports
+extended `PointerEvent`_ messages. Once the server has indicated that
+extended `PointerEvent`_ messages are supported, clients may choose
+between sending normal and extended `PointerEvent`_ messages as needed.
+
+Servers that support this extension must assume that incoming
+`PointerEvent`_ messages are not extended `PointerEvent`_ messages until
+the client indicates that it also supports extended `PointerEvent`_
+messages. From then on forward, the highest bit in the *button-mask* is
+used to indicate whether the message is a normal or an extended
+`PointerEvent`_ message. If an incoming message is a normal
+`PointerEvent`_ message, all buttons represented by the
+*extended-buttons-mask* must be treated as not being pressed.
+
+The bits in *extended-buttons-mask* represent the following buttons
+(see [#extendedmouse]_ for additional context):
+
+========= =============================================================
+Bit       Description
+========= =============================================================
+0         Back
+1         Forward
+2         Reserved
+3         Reserved
+4         Reserved
+5         Reserved
+6         Reserved
+7         Reserved
+========= =============================================================
+
+Note that the back button in *extended-buttons-mask* represents the same
+back button in the *button-mask* for a normal `PointerEvent`_ message.
+
+..  [#extendedmouse] Historically, there has been confusion regarding
+                     which mouse buttons are meant to represent back and
+                     forward. Typically, these buttons are located on
+                     the side of a mouse and are used to navigate back
+                     and forward in applications such as web browsers
+                     and file explorers.
+
 ClientCutText
 -------------
 
@@ -3013,6 +3083,7 @@ Number       Name
 -312         `Fence Pseudo-encoding`_
 -313         `ContinuousUpdates Pseudo-encoding`_
 -314         `Cursor With Alpha Pseudo-encoding`_
+-316         `ExtendedMouseButtons Pseudo-encoding`_
 -317         `Tight Encoding Without Zlib Pseudo-encoding`_
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
 -763 to -768 `JPEG Subsampling Level Pseudo-encoding`_
@@ -4410,6 +4481,21 @@ supported. However some encodings may be unsuitable as they cannot
 include the extra bits that are used for alpha. Also note that the data
 used for the cursor shares state with other rects. E.g. the zlib stream
 for a ZRLE encoding is the same as for data rects.
+
+ExtendedMouseButtons Pseudo-encoding
+------------------------------------
+
+A client which requests the *ExtendedMouseButtons* pseudo-encoding is
+declaring that it is capable of handling additional mouse buttons.
+Clients can notify a server that it supports *ExtendedMouseButtons* by
+sending a `SetEncodings`_ message with the *ExtendedMouseButtons*
+pseudo-encoding.
+
+Servers that support this pseudo-encoding will respond to a client's
+request as soon as possible and send an empty pseudo-rectangle with the
+matching pseudo-encoding. When a client has received the matching
+response from the server, it may chose to start sending extended
+`PointerEvent`_ messages.
 
 Tight Encoding Without Zlib Pseudo-encoding
 -------------------------------------------
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -1937,7 +1937,22 @@ On a conventional mouse, buttons 1, 2 and 3 correspond to the left,
 middle and right buttons on the mouse. On a wheel mouse, each step of
 the wheel is represented by a press and release of a certain button.
 Button 4 means up, button 5 means down, button 6 means left and
-button 7 means right.
+button 7 means right. Button 8 corresponds to the back button, commonly
+found on the side of mice and is used to navigate backwards in certain
+applications.
+
+========= =============================================================
+Bit       Description
+========= =============================================================
+0         Left
+1         Middle
+2         Right
+3         Scroll up
+4         Scroll down
+5         Scroll left
+6         Scroll right
+7         Back
+========= =============================================================
 
 =============== ==================== ========== =======================
 No. of bytes    Type                 [Value]    Description


CendioHalim:
The pseudo-encoding has been registered with [IANA](https://www.iana.org/assignments/rfb/rfb.xhtml#rfb-3) and has the assignment:

 **-316 	ExtendedMouseButtons**

samhed (approved):

rfbproto.rst
```diff
@@ -1674,6 +1674,9 @@ The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
 negotiation of an alternative interpretation for the *x-position*
 and *y-position* fields, as relative deltas.
 
+The `ExtendedMouseButtons Pseudo-encoding`_ extends the mouse button
+functionality by adding an additional byte to the *button-mask*.
+
```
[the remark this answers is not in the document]
CendioHalim: This is already written, the last paragraph at the bottom:
```
Currently, bit 1 in the *extended-button-mask* corresponds to the
forward mouse button, and bit 2 corresponds to the back mouse button.
The remaining 6 bits are reserved.
```

Do you think there should be a similar text added here as well, part of the PointerEvent chapter?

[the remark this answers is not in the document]
berrange: Oh, that's easy to miss at a glance. If we present the information in each chapter as a table rather than prose, I think it'd be more visible. ie in this chapter add

```
=== ==========================
Bit   Button
=== ==========================
0   Left
1   Middle
2   Right 
3   Wheel Up
4   Wheel Down
5   Wheel Left
6   Wheel Right
7   ExtendedPointerEvent Marker
=== =========================
```

And in the latter (new) chapter  add

```
=== ==========================
Bit   Button
=== ==========================
0   Forward
1   Back
2   Reserved 
3   Reserved
4   Reserved
5   Reserved
6   Reserved
7   Reserved
=== =========================
```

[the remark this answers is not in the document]
CendioHalim: That's a good point. It's much easier and quicker to read when presented as a table

rfbproto.rst
```diff
@@ -4088,6 +4092,73 @@ include the extra bits that are used for alpha. Also note that the data
...
+========= =============================================================
+0         Forward
+1         Back
+2         Reserved
+3         Reserved
+4         Reserved
+5         Reserved
+6         Reserved
+7         Reserved
+========= =============================================================
```
berrange, line 4160: This is good, but this commit is missing the companion table under "PointerEvent"

CendioHalim: Added a table to "PointerEvent" as well, but I'm not sure the placement is the best. What do you think?

berrange: It is ok by me.

rfbproto.rst
```diff
@@ -4088,6 +4108,74 @@ include the extra bits that are used for alpha. Also note that the data
 used for the cursor shares state with other rects. E.g. the zlib stream
 for a ZRLE encoding is the same as for data rects.
 
+ExtendedMouseButtons Pseudo-encoding
+------------------------------------
```
CendioOssman, line 4112: Most of this chapter describes the contents of the `PointerEvent` message. I think it's better if those portions are in that message's chapter, as that's where people will look for how it should be handled.

I would suggest keeping just the first paragraph. Possibly also just referring to it as an "extended PointerEvent message" and defining what that means in the other chapter.

rfbproto.rst
```diff
@@ -4088,6 +4108,74 @@ include the extra bits that are used for alpha. Also note that the data
...
+2               ``U16``                         *y-position*
+1               ``U8``                          *extended-button-mask*
+=============== ==================== ========== =======================
+
+While a normal `PointerEvent`_ uses 7 of the 8 available bits, the
+*ExtendedMouseButtons* pseudo-encoding provides support for 8 additional
+bits, for a total of 15 buttons. To allow for 8 additional bits of data
+being transmitted using the same `PointerEvent`_  message, a marker bit
+is used to indicate that there is an extra byte of data in the
+`PointerEvent`_  message. The marker bit is the 7th bit of
```
CendioOssman, line 4137: The 7th bit is bit 6, which is probably not what you mean. :)

Given the system here, I would suggest calling it the highest bit. Least risk of confusion.

rfbproto.rst
```diff
@@ -4088,6 +4108,74 @@ include the extra bits that are used for alpha. Also note that the data
...
+mask.
+
+When decoding the `PointerEvent`_ message with this pseudo-encoding
+enabled, it can initially be decoded as a normal `PointerEvent`_
+message. After decoding the message, the 7th bit in the *button-mask* is
+used to determine whether an additional byte should be read or not. If
+the marker bit is set to one, another byte is read, which contains the
+*extended-button-mask*. The last step in decoding is to set the 7th bit
+of the *button-mask* is to zero.
+
```
CendioOssman, line 4157: You may not need to be this verbose. It's sufficient to clearly state how the messages are expected to look like. Then it's up to the implementation how to send/receive them.

rfbproto.rst
```diff
@@ -4088,6 +4108,74 @@ include the extra bits that are used for alpha. Also note that the data
...
+If the marker bit in  the *button-mask* is set to zero, the message is a
+normal `PointerEvent`_ message and, no further action is needed.
+
+Currently, bit 0 in the *extended-button-mask* corresponds to the
+forward mouse button, and bit 1 corresponds to the back mouse button.
+The remaining 6 bits are reserved and must be set to 0 by the client and
+ignored by the server.
```
CendioOssman, line 4164: As discussed elsewhere, this may need a warning/clarification on what we consider "forward" and "backward".

I suggested a dive in to the USB specification to increase the clarity of this mess. Has that been done?

rfbproto.rst
```diff
@@ -4088,6 +4108,74 @@ include the extra bits that are used for alpha. Also note that the data
...
+Bit       Description
+========= =============================================================
+0         Forward
+1         Back
+2         Reserved
+3         Reserved
+4         Reserved
+5         Reserved
+6         Reserved
+7         Reserved
```
CendioOssman, line 4176: It's best to specify what reserved means for the client and server. I would suggest clients are required to send 0, and servers are required to ignore them.

rfbproto.rst
```diff
@@ -1674,6 +1674,25 @@ The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
 negotiation of an alternative interpretation for the *x-position*
 and *y-position* fields, as relative deltas.
 
+The `ExtendedMouseButtons Pseudo-encoding`_ extends the mouse button
+functionality by adding an additional byte for button states. When
+this pseudo-encoding is enabled, bit 7 in the *button-mask* can be set
+to 1. Clients that don't use this pseudo-encoding must set the marker
+bit to 0.
```
CendioOssman, line 1681: I think we need to consider old/current clients more here.

What happens if a client sends the high bit set, but no ExtendedMouseButtons pseudo-encoding?

And clients supporting ExtendedMouseButtons likely need to be forced to send the high bit 0 until they know the server supports it?

CendioOssman (changes requested):
Looks promising! Some clarifications and we should be good.

RokerHRO:
I'd favour a more flexible and extensible approach, that would also work for so-called "gaming" mice with lots of additional buttons (that might be mapped to keybord shortcts at server side) and for other USB input devices such as styluses or digitizer tablets that are able to report ''strength'' of pressure etc.
 
Maybe a look into the USB HID specification would help, so we don't have to re-invent the wheel, because USB HID already covers other extended "event input" devices.

any1:
> I'd favour a more flexible and extensible approach

Yes, let's also do that, but this is fine as is.

> Maybe a look into the USB HID specification

That's actually a pretty good idea. USB is really just a transport for HID. I know of at least one non-USB use of HID: Miracast passes raw HID events (and descriptors) back to the presenter. This is mainly useful with touch panels.

I look forward to seeing proposals for VNC-HID. :)

CendioOssman:
> I'd favour a more flexible and extensible approach, that would also work for so-called "gaming" mice with lots of additional buttons (that might be mapped to keybord shortcts at server side)

That is something we thought of as well. Turns out, these mice (at least not the ones we have access to) do not send mouse events. They send keyboard events for all those extra buttons. :/

My guess would be because Windows doesn't have a convenient interface for anything more than the back and forward button, so why send hardware events that most applications cannot easily use.

> and for other USB input devices such as styluses or digitizer tablets that are able to report ''strength'' of pressure etc.

That would be nice, but is a much bigger project. Might also require bidirectional signalling to reconfigure things. 

Let's not have those grander ambitions block fixing a common use case like browser interaction.

rfbproto.rst
CendioHalim: > What happens if a client sends the high bit set, but no ExtendedMouseButtons pseudo-encoding

I tested this, using a client that sent a button mask with the high bit set, with other servers (https://github.com/TigerVNC/tigervnc/pull/1711#issuecomment-2145069550).

I should probably test against more implementations to see if there is anything that breaks, but from what I've tested, setting the high bit only affected the RealVNC server, which corresponded with a back button press.

> And clients supporting ExtendedMouseButtons likely need to be forced to send the high bit 0 until they know the server supports it?

Yes, I will add that.

CendioOssman: > I tested this, using a client that sent a button mask with the high bit set, with other servers ([TigerVNC/tigervnc#1711 (comment)](https://github.com/TigerVNC/tigervnc/pull/1711#issuecomment-2145069550)).

What I meant was, what happens when you do this to a server that *supports* this extension? It might expect a second byte that never comes.

rfbproto.rst
CendioHalim: Good point!

rfbproto.rst
CendioHalim: I've updated the spec so that servers that support this extension ignore the marker bit unless the client also supports it. Clients are also required to send the marker bit as 0 if they support the extension, until the server has indicated that it also supports the extension.

rfbproto.rst
CendioHalim: As discussed, I agree it's a good idea. I haven't looked at the USB HID specification yet.

RokerHRO:
Am I right: this protocol extension enables only 2 more mouse buttons (with pre-defined meanings)?

So if a future client wants to support a super fancy mouse with another cool button there is _no way_ for a client to use the "reserved" bits for it nor to negotiate with the server the presence and the meaning of the remaining bits. Another crude "pseudo encoding" hack has to be defined in that case.

CendioHalim:
> Am I right: this protocol extension enables only 2 more mouse buttons (with pre-defined meanings)?

Correct. There are two motivations for this:

1. Simplicity - This is very easy to implement on both the client and server side.
2. QoL - For me personally, this addition is a huge QoL improvement for VNC. While a more general solution would be very cool to have, it would require significant effort to implement. The work needed to implement vs. actual usability, with the current solution, is very high in my opinion.

> So if a future client wants to support a super fancy mouse with another cool button there is _no way_ for a client to use the "reserved" bits for it nor to negotiate with the server the presence and the meaning of the remaining bits. Another crude "pseudo encoding" hack has to be defined in that case.

The point of leaving the remaining bits reserved is to leave room for future additions to this pseudo-encoding. Future additions could be either more buttons, or other hacks for a more general approach (or another pseudo-encoding entirely). 

Having the unused bits reserved as of now should make it easier to preserve backwards compatibility.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
 1               ``U8``                          *button-mask*
 2               ``U16``                         *x-position*
 2               ``U16``                         *y-position*
+1               ``U8``                          *extended-buttons-mask*
```
CendioOssman, line 1671: I don't think we have any precedence of having optional fields like this in a description. I'm worried someone might read this too casually and think this field is always present.

What are your thoughts on keeping the original table, and then having an almost duplicate table showing the extended mode?

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+========= =============================================================
+0         Left
+1         Middle
+2         Right
+3         Wheel-up
+4         Wheel-down
+5         Wheel-left
+6         Wheel-right
+7         ExtendedMouseButtons Marker
+========= =============================================================
```
CendioOssman, line 1689: This table duplicates what's written earlier. It would probably be best to merge things in some way.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+Bit       Description
+========= =============================================================
+0         Left
+1         Middle
+2         Right
+3         Wheel-up
+4         Wheel-down
+5         Wheel-left
+6         Wheel-right
+7         ExtendedMouseButtons Marker
```
CendioOssman, line 1688: This is only well-defined with this new extension. It is probably best to mention that it has no good historical definition. Or did we find any client or server that used it? I would assume it would be the "back" button if used.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+========= =============================================================
 
 The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
 negotiation of an alternative interpretation for the *x-position*
 and *y-position* fields, as relative deltas.
 
+The `ExtendedMouseButtons Pseudo-encoding`_ extends the mouse button
+functionality by adding an additional byte for button states. When
+this pseudo-encoding is enabled, the highest bit in the *button-mask*
+can be set to 1.
```
CendioOssman, line 1698: **must** be set to 1 if this extra byte is used.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+7         ExtendedMouseButtons Marker
+========= =============================================================
 
 The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
 negotiation of an alternative interpretation for the *x-position*
 and *y-position* fields, as relative deltas.
 
+The `ExtendedMouseButtons Pseudo-encoding`_ extends the mouse button
+functionality by adding an additional byte for button states. When
+this pseudo-encoding is enabled, the highest bit in the *button-mask*
```
CendioOssman, line 1697: I'm concerned that this phrasing could be interpreted as it being sufficient to do `SetEncodings` with `ExtendedMouseButtons`.

Perhaps use the phrase "this extension is enabled" and readers will need to look at the other chapter for details?

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+*ExtendedMouseButtons* pseudo-encoding provides support for 8 additional
+bits, for a total of 15 buttons. To allow for 8 additional bits of data
+being transmitted using the same `PointerEvent`_  message, a marker bit
+is used to indicate that there is an extra byte of data in the
+`PointerEvent`_  message. The marker bit is the highest of the
+*button-mask*, where a value of one means that there is an additional
+byte of data to read, and a value of zero means that it is a normal
+`PointerEvent`_ message.
+
+Clients that don't support this pseudo-encoding must set the highest bit
```
CendioOssman, line 1710: We can't retroactively change client behaviour.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+in *button-mask* to 0, and servers that support this extension must
+ignore the marker bit if the client does not support it. If the server
```
CendioOssman, line 1712: "it" here is ambiguous. Should probably explicitly state "this extension".

CendioOssman, line 1712: Is this perhaps a bit harsh? If the client has not signalled support, then shouldn't legacy behaviour persist?

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+does not support this pseudo-encoding, clients must set the marker bit
+to 0.
```
CendioOssman, line 1714: I think it's better to state "Until the client knows the server supports this extension...". It is not just about legacy servers, but also a race until the handshake is complete.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+
+An extended `PointerEvent`_  message is encoded by first setting the
+highest bit of the *button-mask* to one. The *button-mask* will contain
+the button mask for the 7 buttons defined in a normal PointerEvent`_
+message, as well as the marker bit. The rest of the message is sent as a
+normal `PointerEvent`_ message. After sending the *y-position*, an
+additional byte, the *extended-buttons-mask*, is sent containing a one
+byte button mask.
```
CendioOssman, line 1722: Part of this can probably be removed, as it should be clear from the table. In order to keep the section from getting too long.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+
+When decoding the `PointerEvent`_ message with this pseudo-encoding
+enabled, it can initially be decoded as a normal `PointerEvent`_
+message. After decoding the message, the highest bit in the
+*button-mask* is used to determine whether an additional byte should be
+read or not. If the marker bit is set to one, another byte is read,
+which contains the *extended-buttons-mask*.
+
+If the marker bit in  the *button-mask* is set to zero, the message is a
+normal `PointerEvent`_ message and, no further action is needed.
```
CendioOssman, line 1732: It is probably best to explicitly state that any high bits are assumed to be zero when this happens. Otherwise, it becomes unclear what it means to send an extended event with high mouse buttons pressed, followed by a normal event.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+normal `PointerEvent`_ message. After sending the *y-position*, an
+additional byte, the *extended-buttons-mask*, is sent containing a one
+byte button mask.
+
+When decoding the `PointerEvent`_ message with this pseudo-encoding
+enabled, it can initially be decoded as a normal `PointerEvent`_
+message. After decoding the message, the highest bit in the
+*button-mask* is used to determine whether an additional byte should be
+read or not. If the marker bit is set to one, another byte is read,
+which contains the *extended-buttons-mask*.
```
CendioOssman, line 1729: I wonder if this is a bit too step-by-step and hence gets a bit wordy and difficult to read. We should describe the wire format, not how the code should be written.

CendioOssman, line 1729: Somewhere in here, it needs to be explicit that this depends on whether the client has advertised support for the extension or not. Otherwise, a high bit does not mean that there is more data.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+
+If the marker bit in  the *button-mask* is set to zero, the message is a
+normal `PointerEvent`_ message and, no further action is needed.
+
+Currently, bit 0 in the *extended-buttons-mask* corresponds to the
+backward mouse button, and bit 1 corresponds to the forward mouse
```
CendioOssman, line 1735: This is redundant with the table

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+button. The remaining 6 bits are reserved and must be set to 0 by the
+client and ignored by the server.
```
CendioOssman, line 1737: This is good and should still be included. It allows defining those bits at a later date.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+
+Note that on X11 platforms, the back and forward mouse buttons are
+represented by the labels BTN_LABEL_PROP_BTN_SIDE and
+BTN_LABEL_PROP_BTN_EXTRA respectively.
```
CendioOssman, line 1741: Those are rather internal constants and are probably not too useful to most people.

I think it is sufficient to write something generic that there has historically been some confusion on which buttons are Back/Forward. And perhaps instead define them by behaviour. I.e. what they do in applications.

And since it is an implementation hint, it might be best as a note.

rfbproto.rst
```diff
@@ -1668,12 +1668,93 @@ No. of bytes    Type                 [Value]    Description
...
+Bit       Description
+========= =============================================================
+0         Back
+1         Forward
+2         0
+3         0
+4         0
+5         0
+6         0
+7         0
```
CendioOssman, line 1755: `0` suggests that it will always be the only value. "Reserved" is more clear what is going on here.

rfbproto.rst
```diff
@@ -4088,6 +4170,21 @@ include the extra bits that are used for alpha. Also note that the data
...
+
+A client which requests the *ExtendedMouseButtons* pseudo-encoding is
+declaring that it is capable of handling additional mouse buttons.
+Clients can notify a server that it supports *ExtendedMouseButtons* by
+sending a `SetEncodings`_ message with the *ExtendedMouseButtons*
+pseudo-encoding.
+
+Servers that support this pseudo-encoding will respond to a client's
+request by sending an empty pseudo-rectangle with the matching
+pseudo-encoding. When a client has received the matching response from
```
CendioOssman, line 4184: *When* will the server send this? It's best to be explicit.

rfbproto.rst
CendioHalim: Hmm, my reasoning was that you don't **have** to send an extended `PointerEvent` message unless you have to (to save bandwidth or whatever).  The way it is worded now is a bit ambiguous.

Changing it to **must** would make things easier, and the only drawback is that an additional byte is sent per `PointerEvent` message. 

What do you think is best? Forcing all  `PointerEvent`  messages to be extended with this pseudo-encoding, or only forcing it if you want to send the extra byte?

rfbproto.rst
CendioHalim: I tested the following servers:
  * TurboVNC
  * x0vncserver
  * RealVNC
  * libvncserver
    * x11vnc
 
Out of these, only x0vncserver responded to the 7th bit being set, and I was able to navigate "back" on Firefox/Chrome.
 
 I tested the following clients with an Xvnc server that looks at the 7th bit:
   * TurboVNC
   * TightVNC
   * RealVNC
   * Remmina
   * noVNC

Out of these, only RealVNC responded to clicking on the back button.

CendioHalim: I guess this means that the current wording:
```While a normal `PointerEvent`_ uses 7 of the 8 available bits```
is incorrect. The back button is nothing that has ever been a part of the specification, my guess is that the edge cases I've seen are purely coincidental.

File: rfbproto.rst (hunk 8 above)
Comment (CendioHalim) [reply rfbproto.rst]: I thought about this one more time, and I think that the behaviour should be the following:

| Server pseudo-encoding   | Client pseudo-encoding | Behaviour |
| ------------------------------------- | ----------------------------------- | --------------- |
| No                                         | No                                        | Client sends 1 byte, highest bit **can** be set to indicate back button. <br> Server reads 1 byte, **don't** ignore highest bit, **can** interpret highest bit as back button.|
| Yes                                         | No                                        | Client sends 1 byte, highest bit **can** be set to indicate back button. <br> Server reads 1 byte, **don't** ignore highest bit, **can** interpret highest bit as back button.|
| No                                         | Yes                                        | Client sends 1 byte, highest bit **can** be set to indicate back button. <br> Server reads 1 byte, **don't** ignore highest bit, **can** interpret highest bit as back button.|
| Yes                                         | Yes                                        | Client sends 1 byte, highest bit **can** be set to indicate extended `PointerEvent`.  If highest bit is set, another byte **must** be sent.<br> Server reads 1 byte and **must** read highest bit. If highest bit is set, server **must** read another byte.|

This way, we should preserve the old behaviour except for in the case where both the client and server have indicated the pseudo-encoding support, and the client sets the highest bit.

Servers should **only** expect another byte if the client has indicated pseudo-encoding support, and the highest bit is set.

File: rfbproto.rst [hunk 25]
```diff
@@ -1674,6 +1674,67 @@ The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
...
+No. of bytes    Type                 [Value]    Description
+=============== ==================== ========== =======================
+1               ``U8``               5          *message-type*
+1               ``U8``                          *button-mask*
+2               ``U16``                         *x-position*
+2               ``U16``                         *y-position*
+1               ``U8``                          *extended-buttons-mask*
+=============== ==================== ========== =======================
+
+The *ExtendedMouseButtons* pseudo-encoding provides support for 8
```
CendioOssman, line 1693: Technically, 7 additional?

rfbproto.rst
```diff
@@ -1674,6 +1674,67 @@ The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
...
+
+Bit 0 in *extended-buttons-mask* represents the back button on a mouse,
+and bit 1 represents the forward button (see [NOTE2]_ for additional
+context). Bits 2-7 are reserved and must be set to 0 by the client and
+ignored by the server.
+
+========= =============================================================
+Bit       Description
+========= =============================================================
+0         Back
```
CendioOssman, line 1722: I guess we leave the button 8 for the standard message undefined for now?

rfbproto.rst
```diff
@@ -1674,6 +1674,67 @@ The `QEMU Pointer Motion Change Pseudo-encoding`_ allows for the
...
+
+The *ExtendedMouseButtons* pseudo-encoding provides support for 8
+additional button states, giving support for a total of 15 buttons.
+To allow for 8 additional bits of data being transmitted using the same
+`PointerEvent`_ message, a marker bit is used to indicate that there is
+an extra byte of data in the `PointerEvent`_ message. The marker bit is
+the highest bit of the *button-mask*. Clients that support this
+pseudo-encoding must always set the marker bit to zero until the server
+indicates that it supports extended `PointerEvent`_ messages.
+
```
CendioOssman, line 1701: It's still not explicitly stated what happens if you mix standard and extended messages.

CendioOssman (changes requested):

rfbproto.rst
berrange: From the earlier text 

```
 When this pseudo-encoding is enabled, the highest bit in 
the *button-mask* can be set to one, which marks the `PointerEvent`_ 
message as an extended `PointerEvent`_ message"
```

I interpreted it was permitted to mix both standard & extended messages at the clients discretion, since the word `can` was used rather than `must`. Lets make the intention explicit though.

rfbproto.rst
berrange: Hmm, interesting question.   What happens if we have a future situation where some VNC server/client that's unaware of this extension, defines a mapping for button 8 in the standard message. If it later wants to use the extended message, it can't use button 8 in the standard message anymore, as that bit is used to indicate that extended message. We would nee to re-map it into the extended message I guess ?

CendioOssman: I think that ship has sailed, given that RealVNC and x0vncserver already have some behaviour for that bit. So it's more that we haven't explicitly documented it.

berrange: Hmm, awkward. I can't help thinking that rather than having the encoding alter the semantics of the PointerEvent message, it would be simpler if we requested assignment of a "ExtendedPointerEvent" message, and then just defined its button-mask as a `u16`. That way we wouldn't have to assign magic semantics to bit 8, which clash with pre-existing usage of bit 8.

CendioOssman: Indeed. But the namespace for messages is a lot smaller than for encodings. So they idea here was to avoid an allocation if possible.

CendioOssman: And it's not like that button is somehow lost. It just gets shifted up to the lowest bit in the extended byte.

rfbproto.rst
CendioHalim: > I interpreted it was permitted to mix both standard & extended messages at the clients discretion, since the word `can` was used rather than `must`. Lets make the intention explicit though.

That was my reasoning, but yeah this could be described more explicitly.

rfbproto.rst
CendioHalim: > Hmm, interesting question. What happens if we have a future situation where some VNC server/client that's unaware of this extension, defines a mapping for button 8 in the standard message. If it later wants to use the extended message, it can't use button 8 in the standard message anymore, as that bit is used to indicate that extended message. We would nee to re-map it into the extended message I guess ?

> And it's not like that button is somehow lost. It just gets shifted up to the lowest bit in the extended byte.

Indeed, button 8 would be assigned to the first bit in the extended mask.

The only thing that we break with this pseudo-encoding is that clients that want to enable this feature are **not** allowed to send bit 8 when sending normal `PointerEvent` messages. The damage is quite minimal in my opinion, as there are barely any servers/clients that use bit 8 anyway and the bit's meaning is not defined in the protocol.

CendioHalim: > I guess we leave the button 8 for the standard message undefined for now?

I don't feel confident that we can/should define button 8, undefined is probably best.

rfbproto.rst
CendioHalim: Or perhaps using the word `may`, which is used many times in the specification already? It also matches [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) more closely.

rfbproto.rst
CendioHalim: Hmm, depends. If we specify bit 8 in the normal message as undefined, then we do technically get 8 more bits to play with.

rfbproto.rst
CendioHalim: > Out of these, only RealVNC responded to clicking on the back button.

My mistake, what I saw was that RealVNC server `vncserver-x11`, (not `vncserver-virtual`) responded on clicking the back button.

CendioOssman (changes requested):
Looks pretty good. Some clarifications with the last changes could be good, then we can merge.

CendioOssman (approved):

PR #58, opened 2024-07-29
Clarify Open H.264 handling of oversized frames

any1:
This clarifies how clients should treat frames that are either larger or smaller than the rectangle to which they belong.

Hardware encoders are often limited in that frame sizes need to be multiples of certain numbers such as 16 or 32. It is therefor advantageous to allow servers to submit oversized frames, expecting the client to crop them.

See: https://github.com/TigerVNC/tigervnc/pull/1785

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -3526,6 +3526,11 @@ into multiple packets. The *data* field must always contain 0 or more
 whole frames. The client must first decode all the received frames
 in a single message, and then display the result.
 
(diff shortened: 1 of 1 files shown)


CendioOssman (changes requested):
Looks good! Thanks!

If you could please add a signed-off-by line, and I'll get it merged.

any1:
Signed off.

CendioOssman (approved):
Thanks!

PR #64, opened 2024-09-24
Document Tight Encoding Without Zlib extension

dcommander:
This extension has been used unofficially in TurboVNC since 2008, in order to support a Tight-based encoding method that compresses almost as well as Hextile (4-6x relative to Raw) while using about half as much CPU time as Hextile.  The original purpose of the extension was to support older SPARC CPUs that couldn't keep up with JPEG or zlib decompression on a gigabit network.  However, it has found a new purpose on loopback connections to virtual machines, as well as in enterprise environments that run a VNC viewer inside of Citrix or Remote Desktop and want to minimize re-encoding overhead while still reducing server-area network usage.

Even with zlib-ng (which has SIMD acceleration and numerous other optimizations), bypassing zlib altogether results in about a 10% speedup in average Tight encoding performance relative to zlib level 0 (assuming that JPEG subencoding is disabled.)  With the unaccelerated zlib implementation, the speedup is about 60% for zlib 1.2.7 and 15% for zlib 1.2.11.

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -2720,6 +2720,7 @@ Number       Name
 -312         `Fence Pseudo-encoding`_
 -313         `ContinuousUpdates Pseudo-encoding`_
 -314         `Cursor With Alpha Pseudo-encoding`_
+-317         `Tight Encoding Without Zlib Pseudo-encoding`_
 -412 to -512 `JPEG Fine-Grained Quality Level Pseudo-encoding`_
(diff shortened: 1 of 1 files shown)


CendioOssman (changes requested):
Looks good to me, except that some lines seem to be longer than the 72 columns we try to maintain. Could you have a look at wrapping those lines?

dcommander:
Should be good now.

CendioOssman (approved):
Looks good! Thanks!

PR #66, opened 2024-12-11
Define semantics for colour map after VMWare Display Mode Change psuedo-encoding

berrange:
This originates with a bug I received against GTK-VNC, where we fail to preserve the colour map entries when handling any messages which trigger re-creation of the local framebuffer - either a desktop resize, or a pixel format change.

In fixing this I sought to understand when we should & should not preserve colour map entries. I observed the spec says we should NOT preserve colour map entries on SetPixelFormat, but failed to say anything about "VMVare Display Mode Change" & I think the semantics of the latter should match the former.

~Note, QEMU is not in compliance with these new documented semantics, as while it sends "SetColorMapEntries" after receiving a "SetPixelFormat" message, it fails to do the same after sending the "VMware Display Mode Change" psuedo-encoding.~

~Fixing QEMU to follow the new spec, should be backwards compatible with existing clients:~

* ~If a client DID NOT previously reset its view of the colour map. It would carry on with old (outdated) colourmap. Seeing a new SetColorMapEntries message is something the client should already be ready to handle. With QEMU fixed, a client using stale colour map info would now have correct colour map info. Functionally better.~
 
* ~If a client DID previously reset its view of the colour map. It would be lacking the ability to render framebuffer updates due to lack of SetColorMapEntries message from QEMU. With QEMU fixed, a previously broken client would start working.~

### DIFF
--- a/rfbproto.rst
+++ b/rfbproto.rst
@@ -4734,23 +4734,8 @@ Clients supporting multi-head layouts must wait for this following
 *ExtendedDesktopSize* rectangle to determine the new layout.
 
 The *width*  and *height* of the pseudo-rectangle specify the new width
-and height of the display. The rest of the format is described below:
-
-=============== ==================== =======================
-No. of bytes    Type                 Description
-=============== ==================== =======================
-1               ``U8``               *bits-per-sample*
-1               ``U8``               *depth*
-1               ``U8``               *color*
-1               ``U8``               *true-color*
-2               ``U16``              *max-red*
-2               ``U16``              *max-green*
-2               ``U16``              *max-blue*
-1               ``U8``               *red-shift*
-1               ``U8``               *green-shift*
-1               ``U8``               *blue-shift*
-3                                    padding
-=============== ==================== =======================
+and height of the display, followed by ``PIXEL_FORMAT`` as described in
+`ServerInit`_.
 
 VMware Virtual Machine State Pseudo-encoding
 --------------------------------------------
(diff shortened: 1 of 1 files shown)


berrange:
> Note, QEMU is not in compliance with these new documented semantics, as while it sends "SetColorMapEntries" after receiving a "SetPixelFormat" message, it fails to do the same after sending the "VMware Display Mode Change" psuedo-encoding.

Actually that's incorrect. QEMU trivially complies because when it sends "VMWare Display Mode Change" it is hardcoded to always set true-color-flag=True. The only scenario QEMU will ever use a colourmap is when asked in a SetPixelFormat message.

berrange:
@CendioOssman any thoughts on this one ?

rfbproto.rst
```diff
@@ -4734,23 +4734,24 @@ Clients supporting multi-head layouts must wait for this following
 *ExtendedDesktopSize* rectangle to determine the new layout.
 
 The *width*  and *height* of the pseudo-rectangle specify the new width
-and height of the display. The rest of the format is described below:
+and height of the display, followed by ``PIXEL_FORMAT`` as described in
+`ServerInit`_:
```
CendioOssman, line 4744: Should we perhaps remove this duplication and readers will have to check the other chapter?

CendioOssman (approved):
Sorry, this got lost in the todo pile.

Changes look good to me!

One further improvement could be to completely remove the duplicated table. What do you think?

rfbproto.rst
berrange: Pushed new version with that change made

CendioOssman (approved):
Nice!

PR #75, opened 2026-09-19
Document the UltraVNC security type

BodholdtSpeedTesting:
Security type 17 is registered as "Ultra" and the document says nothing about what happens once a client selects it. It turns out not to be an authentication method at all: it starts a second security handshake, and the server sends a list of security types of its own to choose from. A client that selects 17 and then waits for a challenge waits indefinitely.

This adds a section describing the exchange, moves 17 into the list of types the document defines, and records two numbers seen alongside it.

### Where it came from

I watched a running UltraVNC server on my own network with a small probe that selects one security type and prints whatever comes back. I did not read the implementation — I was writing a clean-room client that cannot touch GPLv3 source — so the field names are mine, and I have tried to separate what was observed from what is guessed.

```
select 17, VNC password only    ->  ff ff ff ff 01 02
select 17, MS-Logon II enabled  ->  ff ff ff ff 02 76 71
then answer 02                  ->  16 bytes (a VNC Authentication challenge)
then answer 113, not offered    ->  no reply at all
```

So: four bytes, a count, then that many security types, and the client picks one. Answering a type the server had not listed produced silence rather than a SecurityResult of failed.

Note this is not a new extension and needs no number allocation — 17 is already registered. It is only the handshake behind it that was undocumented.

### What I have not claimed

I do not know what the `0xFFFFFFFF` prefix is. It was identical in every exchange, which is not the same as understood. It could be a sentinel, a version, or a count of a structure that is always empty; `Tight Security Type` opens with a *number-of-tunnels*, so that shape would not be surprising. The section says it can be treated as four opaque bytes and leaves it there.

I have also not claimed 118 is MS-Logon III. The evidence lines up — 32 bytes is X25519 key length, it appears only with the newer MS-Logon setting, and the UltraVNC 1.8.x release notes describe MS-Logon III as X25519 ECDH with AES-256-GCM — but lining up is not confirmation, and the table says so.

If someone who has read the implementation can settle either point, the section will be better for it.

### Checks

Builds with no new docutils warnings. Lines are within 72 characters, the file stays pure ASCII, and the tables use the simple form at the existing widths.

### Happy to change things

I can reword anything, drop the "Other numbers seen in use" table if it does not belong here, or capture more exchanges against other server configurations if that would help. I am glad to move the discussion to the mailing list if that is preferred for a change of this size.


rfbproto.rst
```diff
@@ -1415,6 +1415,98 @@ No. of bytes      Type          Description
...
+----------------------
+
+Security type 17 is not an authentication method. Selecting it starts a
+second security handshake: the server sends a list of security types of
+its own, and the client picks one of those instead.
+
+Everything in this section was observed from a running UltraVNC server.
+The implementation was not consulted, so the field names below are
+descriptive rather than official, and the first four bytes are not
+explained.
```
berrange: This isn't really appropriate for a protocol specification IMHO. The spec needs precision, and given the impl is OSS , that really needs someone to look at the implementation to get intended semantics correct.

rfbproto.rst
```diff
@@ -1415,6 +1415,98 @@ No. of bytes      Type          Description
...
+
+followed by *number-of-types* repetitions of the following:
+
+=============== =============================== =======================
+No. of bytes    Type                            Description
+=============== =============================== =======================
+1               ``U8``                          *security-type*
+=============== =============================== =======================
+
+The *prefix* was ``0xFFFFFFFF`` in every exchange seen. It may be a
```
berrange: This is wrong - there's no such thing as a "prefix" in the security type. The 0xFFFFFFF value seen is an extension to the "SecurityResult" part of the protocol:

  https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#securityresult

which can be seen defined here:

  https://github.com/ultravnc/UltraVNC/blob/main/rfb/rfbproto.h

```// adzm 2010-09 - rfbUltraVNC or other auths may send this to restart authentication (perhaps over a now-secure channel)
#define rfbVncAuthContinue 0xFFFFFFFF
```


The rfbUltraVNC security type is essentially a no-op - it is just a mechanism for the client to identify the remote server as being UltraVNC, which then implicitly enables the Ultra VNC file transfer protocol extension. It sends rfbVncAuthContinue as the SecurityType, which re-starts the standard authentication process.

rfbproto.rst
```diff
@@ -1415,6 +1415,98 @@ No. of bytes      Type          Description
...
+        tested, including with authentication disabled. Selecting it
+        returns a *SecurityResult* of failed with the reason
+        "authentication rejected", immediately and with no handshake.
+        Its purpose was not determined.
+118     Offered only when MS-Logon is enabled together with the "New
+        MS-Logon" option. Selecting it returns 32 bytes of high entropy
+        data. That is the length of an X25519 public key, and the
+        UltraVNC 1.8.x release notes describe an "MS-Logon III" using
+        "X25519 ECDH + AES-256-GCM via libsodium". The two are
+        consistent, but this was not confirmed.
```
berrange: This is all redundant - it is just re-documenting the standard security handshake.

118 is simply a new security type, a variant of https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#mslogonii-authentication  which uses X25519. This has no relation to the UltraVNC security type, and needs to be documented in its own right.

